PR #8, opened 2021-03-29
Update and delete

tildeee:


update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+![Screenshot of Postman featuring a request of PUT /books/1 and a response of 200 with a success message](../assets/update-and-delete/update-and-delete_update-request.png)
+
+We can send another request to `GET /books/1` to confirm the updates worked.
+
+![Screenshot of Postman featuring a request of GET /books/1 and a response updated book data](../assets/update-and-delete/update-and-delete_update-confirm-get.png)
+
+### Manually Testing with `psql`
+
+We can also check if our book was updated in the database using `psql`!
```
anselrognlie, line 163: flesh this out a bit as in the last lesson (at least for this first one)
```suggestion
We can also check if our book was updated in the database using `psql`, just as we did when creating new records!

Let's try it out on our own! We can refer back to the Intro to SQL lessons if we need a reminder about how to write a query like this.
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+##### !question
+
+Think about the "Updating a Book Endpoint."
+
+Check off all the topics that we've briefly touched on so far.
+
+##### !end-question
+##### !options
+
+* Predicted the HTTP response, request, and logic for this endpoint
```
anselrognlie, line 179: match lesson changes
```suggestion
* Planned the HTTP response, request, and logic for this endpoint
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+* Refactored our `"/books/<book_id>"` endpoint to accept `PUT` requests
+* Used `Book.query.get(book_id)` to get an instance of `Book` matching `book_id`
+* Refactored our code to conditionally check the request's method
+* Read the HTTP request body using `request.get_json()`
+* Saved the instance of `Book` in the database
```
anselrognlie, line 184: reflect the actual action
```suggestion
* Updated the instance of `Book` in the database
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+* Returned a response
+* Tested this request in Postman
+
+##### !end-options
+##### !answer
+
+* Predicted the HTTP response, request, and logic for this endpoint
```
anselrognlie, line 191: match the question revision
```suggestion
* Planned the HTTP response, request, and logic for this endpoint
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+* Refactored our `"/books/<book_id>"` endpoint to accept `PUT` requests
+* Used `Book.query.get(book_id)` to get an instance of `Book` matching `book_id`
+* Refactored our code to conditionally check the request's method
+* Read the HTTP request body using `request.get_json()`
+* Saved the instance of `Book` in the database
```
anselrognlie, line 196: match the question revision
```suggestion
* Updated the instance of `Book` in the database
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+##### !end-answer
+### !end-challenge
+<!-- prettier-ignore-end -->
+
+### !callout-warning
+
+## What About Error Handling?
+
+There are many cases that weren't covered in this lesson, even though they are relevant to creating and reading `Book`s. For example, what happens if we make a `GET` request to `/books/this-book-doesnt-exist`? We are intentionally not covering these cases at the moment, to limit this lesson. However, hypothesize and research how to handle erroneous HTTP requests. Follow your curiosity!
```
anselrognlie, line 208: match the verb
```suggestion
There are many cases that weren't covered in this lesson, even though they are relevant to creating and reading `Book`s. For example, what happens if we make a `PUT` request to `/books/this-book-doesnt-exist`? We are intentionally not covering these cases at the moment, to limit this lesson. However, hypothesize and research how to handle erroneous HTTP requests. Follow your curiosity!
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+### !end-callout
+
+## Deleting a Book Endpoint: Preparation
+
+Let's consider how to accomplish this feature:
+
+> As a client, I want to send a request to delete one existing book and get a success response, so that I know the API deleted the book data.
+
+### Predict HTTP Requests, Responses, and Logic
```
anselrognlie, line 218:
```suggestion
### Planning HTTP Requests, Responses, and Logic
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+Let's consider the endpoint to create a book:
```
anselrognlie, line 220: prompt for verb and endpoint consideration
```suggestion
We should think about the typical HTTP verb and endpoint used for requests that delete a particular model record.

For this feature, we should make a `DELETE` request to the `/books` path, and we'll need to include the `id` of the record to delete as part of the endpoint.
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+| HTTP Method | Endpoint   |
+| ----------- | ---------- |
+| `DELETE`    | `/books/1` |
+
+No request body.
```
anselrognlie, line 226: details about why we don't need the request body
```suggestion
`DELETE` requests do not generally include a request body, so no additional planning around the request body is needed.
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+This assumes that there is at least a `book` table with the following row:
+
+| `id` | `title`                        | `description`                                     |
```
anselrognlie, line 230: formatting
```suggestion
| <div style="min-width:40px;">`id`</div> | `title`                | `description`                               |
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+| ---- | ------------------------------ | ------------------------------------------------- |
+| `1`  | `Updated Fictional Book Title` | `This fantasy novel has robots and pirates, too.` |
+
+The response we want to send back is:
```
anselrognlie, line 234: discussion of the response
```suggestion
When the record is successfully updated, we should return the appropriate status code, which is `200 OK`. Some APIs return the the record that was just deleted as the response body, but for simplicity, we will return a brief status message.
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+| Response Status | Response Body                  |
+| --------------- | ------------------------------ |
+| `200 OK`        | `Book #1 successfully deleted` |
+
+During this method, we will need to:
```
anselrognlie, line 240: typical transition text
```suggestion
Now that we have an idea of what our endpoint should look like, we can turn our attention to how to implement it.

Our endpoint will need to:
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+1. Read the new, updated book data from the HTTP request
+1. Find the instance of `Book` that has the existing book data
+1. Delete this instance of `Book`
+1. Save this change in the db
+1. Send back a response
```
anselrognlie, line 246: call out using the route param and looking it up in the db
characterize the `delete` call as informing the database of our intent to delete, prior to the commit call

```suggestion
1. Read the `book_id` in the request path
1. Retrieve the `Book` instance with the matching `book_id` from the database
1. Tell the database to delete this instance of `Book`
1. Save this change in the database
1. Send back a response
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+## Deleting a Book Endpoint: Code
+
+Let's expand our `book` endpoint.
+
+Let's look at this example code for our delete feature.
```
anselrognlie, line 252: similar updates as above
```suggestion
This endpoint uses the same path as our existing parameterized route, `"/<book_id>"`, so we can refactor and expand on the same function.

Let's modify our endpoint code to support the delete feature.
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+```python
+@books_bp.route("/<book_id>", methods=["GET", "PUT", "DELETE"])
+def book(book_id):
+    book = Book.query.get(book_id)
```
anselrognlie, line 257: as mentioned above, we should consider not having a variable shadow the function it's defined in

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+    if request.method == "GET":
+        # ... existing code for getting a single book
+    elif request.method == "PUT":
+        # ... existing code for updating a single book
+    elif request.method == "DELETE":
+        db.session.delete(book)
+        db.session.commit()
+        return Response(f"Book #{book.id} successfully deleted", status=200)
```
anselrognlie, line 266:
```suggestion
        return make_response(f"Book #{book.id} successfully deleted")
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+```
+
+| <div style="min-width:250px;"> Piece of Code </div> | Notes                                                                                                                                                 |
+| --------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `methods=["GET", "PUT", "DELETE"]`                  | We must update this endpoint so it accepts `DELETE` requests                                                                                          |
+| `book = Book.query.get(book_id)`                    | This feature will need to find our matching `Book` instance, so we still need to use `book`                                                           |
```
anselrognlie, line 272: remind what book is and what this line does to fill it
```suggestion
| `book = Book.query.get(book_id)`                    | This feature will need to find our matching `Book` instance, so we still need the lookup code using `book_id` and store the found book in `book`                                                           |
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+| `elif request.method == "DELETE":`                  | We can add a new conditional branch for `DELETE` requests                                                                                             |
+| `db.session.delete(book)`                           | We can use SQLAlchemy's functions to delete our `book` with `db.session.delete(book)`                                                                 |
```
anselrognlie, line 274: explain how this is notifying sqlalchemy of the pending deletion
```suggestion
| `db.session.delete(book)`                           | We can use SQLAlchemy's functions to tell the database to prepare to delete our `book` with `db.session.delete(book)`                                                                 |
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+| `db.session.commit()`                               | We use this function to actually apply our database changes                                                                                           |
+| `return Response(...)`                              | This is one of many ways to return an HTTP response                                                                                                   |
```
anselrognlie, line 276:
```suggestion
| `return make_response(...)`                              | This is one of many ways to return an HTTP response. Since we didn't supply a status code, Flask will default to `200 OK`.                                                                                                   |
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+| `f"Book #{book.id} successfully deleted"`           | We can still access `book.id` because the variable itself is still in scope in our app, even if the `book` _was_ successfully deleted in the database |
```
anselrognlie, line 277: explicitly state that retrieved book instance isn't a live copy attached to changes in the database
```suggestion
| `f"Book #{book.id} successfully deleted"`           | We can still access `book.id` because the variable itself is still in scope in our app, even if the `book` _was_ successfully deleted in the database. The object stored in memory and referenced by `book` does not automatically update in response to changes in the database. |
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+Then, let's make our `DELETE` request to `/books/1` and check the response.
+
+![Screenshot of Postman featuring a request of DELETE to /books/1 and a response of 200 with a success message](../assets/update-and-delete/update-and-delete_delete-request.png)
+
+Afterwards, let's even make a `GET` request to `/books`. We see that there are no more books that exist! Our book was successfully deleted.
+
+![Screenshot of Postman featuring a request of GET to /books and a response of 200 with an empty array](../assets/update-and-delete/update-and-delete_delete-get-books-empty.png)
+
+We could also use `psql` to check this, too.
```
anselrognlie, line 295: similar (but smaller) expansion as above
```suggestion
### Manually Testing with `psql`

We can also check if our book was deleted from the database using `psql`.

Let's try it out on our own!
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+##### !question
+
+Think about the "Deleting a Book Endpoint."
+
+Check off all the topics that we've briefly touched on so far.
+
+##### !end-question
+##### !options
+
+* Predicted the HTTP response, request, and logic for this endpoint
```
anselrognlie, line 311:
```suggestion
* Planned the HTTP response, request, and logic for this endpoint
```

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+* Refactored our code to conditionally check the request's method
+* Deleted the `book` from the database
+* Committed these changes in the database
+* Returned a response
+* Tested this request in Postman
+
+##### !end-options
+##### !answer
+
+* Predicted the HTTP response, request, and logic for this endpoint
```
anselrognlie, line 323:
```suggestion
* Planned the HTTP response, request, and logic for this endpoint
```

update-and-delete/activity-solar-system-api-update-delete.md
```diff
@@ -0,0 +1,57 @@
+# Activity: Solar System API: Update, Delete
```
anselrognlie, line 1: Ah. Now I see why the previous incarnation of solar system switched drivers daily!

update-and-delete/activity-solar-system-api-update-delete.md
```diff
@@ -0,0 +1,57 @@
...
+- Meet with your pair/group 👋
+- Decide which members will take the driver and navigator roles. This should be a switch compared to the last activity.
+
+The new driver may need to setup their database. Follow these steps:
+
+1. Pull down all new Git commits
+1. Activate the virtual environment
+1. Install the dependencies
+1. Create the database `solar_system_development`
+1. Create the `planet` table with the columns `id`, `name`, and `description`, and the one additional column
```
anselrognlie, line 26: this should be unnecessary. running the upgrade should apply the migration, creating the table
```suggestion
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
+# 404s and More Queries
+
+## Goals
+
+Our goal for this lesson is to apply our learnings about edge cases from RESTful API design. We should walk away from this lesson understanding how to write more robust Flask code.
+
+This lesson covers:
+
+- Handling when a resource isn't found
+- Exploring more query functionality available by Flask-SQLAlchemy
```
anselrognlie, line 10:
```suggestion
- Exploring more query functionality made available by Flask-SQLAlchemy
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Hello Books API
+
+### Before This Lesson
+
+This lesson uses the Hello Books API.
+
```
anselrognlie, line 17: details often need a br to push them down.
```suggestion

<br />

```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+</details>
+
+## `GET`ting a Missing Book: Preparation
+
+Let's consider how to accomplish this feature:
+
+> As a client, I want to send a request trying to get one non-existing book and get a 404 response, so I know that the book resource was not found.
+
+### Predict HTTP Requests, Responses, and Logic
```
anselrognlie, line 47:
```suggestion
### Planning HTTP Requests, Responses, and Logic
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+Let's consider the endpoint to get a single book:
```
anselrognlie, line 49: prompts for why we used this endpoint
```suggestion
This feature is a variation on our existing endpoint that reads a record, but it uses an invalid id. As a result, we can use the following verb and endpoint combination.
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+| HTTP Method | Endpoint          |
+| ----------- | ----------------- |
+| `GET`       | `/books/99999999` |
+
+No request body.
```
anselrognlie, line 55: more explicit reasoning for lack of request body
```suggestion
As with our existing `GET` requests, we do not send a request body.
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+The response we want to send back is:
```
anselrognlie, line 57: reasoning about the response
```suggestion
We want to inform the client that the record was not found. The most appropriate response status code is `404 Not Found`. We could return an error message that echoes back the id the client submitted, but for simplicity, we'll just return the status code and leave the body empty.
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+| Response Status | Response Body |
+| --------------- | ------------- |
+| `404 Not Found` | -             |
+
+During this method, we will need to:
```
anselrognlie, line 63: slight variation on the transitional phrase we've been using
```suggestion
Now that we have planned out the endpoint behavior, we can turn our attention to how to implement it.

Our endpoint will need to:
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+1. Read the `book_id` in the request path
+1. Retrieve the book with the matching `book_id` from the db
+1. Discover that there is no matching book with `book_id` in the db
+1. Send back a response
+
+### Working More With `query`
+
+We've been retrieving our `Book` instance with the line `Book.query.get(book_id)`. What does this method do when there is no matching book?
+
+When `Model.query.get(primary key)` doesn't find a matching record, it returns `None`!
```
anselrognlie, line 74: primary_key to be more pythonic
```suggestion
When `Model.query.get(primary_key)` doesn't find a matching record, it returns `None`!
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+## `GET`ting a Missing Book: Code
+
+For this feature, we must return to our `GET` `/<book_id>` endpoint and re-work it.
```
anselrognlie, line 78: reiterate why we're going back to the same route
```suggestion
This endpoint uses the same path as our existing parameterized route, `"/<book_id>"`, so we can re-work it to add our new behavior.
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+Let's look at this example code for one strategy:
```
anselrognlie, line 80: invite the reader to perform this modification even though we're going to refactor
```suggestion
Let's modify our endpoint code to report a not found status when trying to `GET` a book.
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+```python
+@books_bp.route("/<book_id>", methods=["GET", "PUT", "DELETE"])
+def book(book_id):
+    book = Book.query.get(book_id)
+
+    if request.method == "GET":
+        if book == None:
+            return Response("", status=404)
```
anselrognlie, line 89: when checking against `None`, `is` is more idiomatic than `==`. What I'd really like to  use is `if not book:` but that seems like it would need a callout that would be more distracting than helpful.

also replace Response with call to helper
```suggestion
        if book is None:
            return make_response("", 404)
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+            "description": book.description
+        }
+    # ... existing code for updating a single book
+    # ... existing code for deleting a single book
+```
+
+| <div style="min-width:250px;"> Piece of Code </div> | Notes                                                                                       |
+| --------------------------------------------------- | ------------------------------------------------------------------------------------------- |
+| `if book == None:`                                  | This checks if `Book.query.get(book_id)` returned `None` because there was no matching book |
+| `return Response("", status=404)`                   | We sent back an empty string as our empty response body                                     |
```
anselrognlie, line 102: reflect the above changes in the explanation
```suggestion
| `if book is None:`                                  | This checks if `Book.query.get(book_id)` returned `None` because there was no matching book |
| `return make_response("", 404)`                   | We sent back an empty string as our empty response body and set the status code to `404 Not Found`                                     |
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Countless Possible Refactorings
+
+Handling a missing book and sending back a 404 happens to be the same response we want for all three endpoints:
+
+- `GET` `/books/<book_id>`
+- `PUT` `/books/<book_id>`
+- `DELETE` `/books/<book_id>`
+
+We can consider refactoring and then testing our API to this code:
```
anselrognlie, line 120: briefly mention the alternative to provide justification for this change.
```suggestion
While we _could_ duplicate the `None` check we added to the `GET` branch in each of the other cases, a simpler approach is to handle the `None` check before even checking for the verb. This lets us handle all three cases with a single piece of logic!

Let's make the change and then test our API, making sure each verb case works as expected!
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+```python
+@books_bp.route("/<book_id>", methods=["GET", "PUT", "DELETE"])
+def book(book_id):
+    book = Book.query.get(book_id)
+    if book == None:
+        return Response("", status=404)
```
anselrognlie, line 127: equivalent changes as above
```suggestion
    if book is None:
        return make_response("", 404)
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+        }
+    # ... existing code for updating a single book
+    # ... existing code for deleting a single book
+```
+
+### !callout-info
+
+## Even More Refactoring
+
+We could refactor our code to a number of different solutions that are more elegant. Feel free to refactor!
```
anselrognlie, line 143: elegant code is pretty subjective, and depends on what aspects you value. Maybe just leave it open ended?
```suggestion
There is no single correct way to structure our code. We could refactor our code to any number of different solutions. Feel free to refactor!
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+### !end-challenge
+<!-- prettier-ignore-end -->
+
+## More About Queries
+
+[There's more to explore with querying](https://flask-sqlalchemy.palletsprojects.com/en/2.x/queries/#querying-records)! Let's touch on two topics.
+
+### Filter By
+
+We can use the method `filter_by()` in order to filter our search query. We can give keyword arguments to describe the attribute and value we're filtering on. Consider this example that filters `Book`s by title.
```
anselrognlie, line 186: dangling preposition
```suggestion
We can use the method `filter_by()` in order to filter our search query. We can give keyword arguments to describe the attribute and value on which we're filtering. Consider this example that filters `Book`s by title.
```

more-flask-queries/404s-and-more-queries.md
```diff
@@ -0,0 +1,237 @@
...
+
+```python
+Book.query.filter_by(title="Fictional Book Title")
+```
+
+### !callout-info
+
+## Way More to Explore
+
+There's way more to explore in filtering that is uncovered in this curriculum, so drive your own learning!
```
anselrognlie, line 196: my brain _really_ wants "way more than" not "way more that", so while what was there is technically fine, I'd suggest this change.

```suggestion
There's way more to explore in filtering than is covered in this curriculum, so drive your own learning!
```

anselrognlie (commented):
comments for 404 and more

more-flask-queries/404s-and-more-queries.md
tildeee: I think `if not book` is totally reasonable at this point in the curriculum?

```suggestion
        if not book:
            return make_response("", 404)
```

more-flask-queries/404s-and-more-queries.md
tildeee: If we're excited about `if not book`:

```suggestion
| `if not book:`                                  | This checks if `Book.query.get(book_id)` returned `None` because there was no matching book |
| `return make_response("", 404)`                   | We sent back an empty string as our empty response body and set the status code to `404 Not Found`                                     |
```

more-flask-queries/404s-and-more-queries.md
tildeee: Note for me/us to consider if the phrase "None check" is appropriate if we go the `if not book` way

update-and-delete/update-and-delete.md
tildeee: Just want to explicitly say that I like this a lot and I'm learning form it!

update-and-delete/update-and-delete.md
tildeee: This specific suggestion is helping me remember to continue to be repetitive in this curriculum (which I put a lot of effort into in Unit 1, but probably slipped on during Unit 2. Good to refresh myself for Unit 3!)

update-and-delete/update-and-delete.md
```diff
@@ -0,0 +1,355 @@
...
+
+### Manually Testing with Postman
+
+We can use Postman to check our update functionality.
+
+It may help to first make a request to `GET /books/1` before updating it.
+
+![Screenshot of Postman featuring a request of GET /books/1 and a response of 200 with book data](../assets/update-and-delete/update-and-delete_update-precheck-get.png)
+
+Then, make a request to `PUT /books/1` with our proper JSON request body.
```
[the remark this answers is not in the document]
tildeee: This is such a small suggestion, I recognize I might be fixating on it, but I don't feel like this is necessary, and I would love reassurance that this will help the student experience!

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+From this example, we can observe:
+
+| Original character | Encoding character |
+| ------------------ | ------------------ |
+| ` ` (space)        | `%20`              |
+| `:`                | `%3A`              |
+| `!`                | `%21`              |
+
+Often, the work of encoding happens _through the client_, not through the server. However, this knowledge may be helpful when constructing URLs with query strings for testing!
```
anselrognlie, line 91: expand the guidance around when url encoding might be necessary
I really wanted to spell out that the replacement sequence is a percent followed by the ascii code of the character to replace encoded as 2 hexadecimal digits, but the curriculum only mentions ascii in passing once, and NEVER mentions hexadecimal.

More lightning talk candidates? (ascii/unicode, hexadecimal/number bases)

```suggestion
If we are using a program like Postman or a web browser to send requests, they will do their best to encode our URLs for us. If we are writing code to contact an endpoint, we may need to encode parts of the URL ourselves, using functionality built into Python.

We will try to avoid characters that require encoding in our URLs and query parameters. However, this knowledge may be helpful when constructing URLs with query strings for testing!
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+### Typical Use Cases
+
+Query params provide extra information to an HTTP request.
+
+HTTP response bodies and query params are both responsible for adding data to a request, and can both accomplish similar things. The exact difference in usage will always depend on your project, context, and needs.
```
anselrognlie, line 97: also mention route parameters
```suggestion
HTTP request bodies, route parameters, and query params are all responsible for adding data to a request, and can accomplish similar things. The exact difference in usage will always depend on your project, context, and needs.
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+However, typically:
+
+- HTTP response bodies contain data that should be uploaded to the server
+- Query params contain data that describe the nature of the request
```
anselrognlie, line 102: we don't need both however and typically
include info about route parameters
describe the nature seems vague. suggest presenting as customizing the request
```suggestion
In the style of web API we are building, typically:

- HTTP request bodies contain data that should be uploaded to the server
- Route parameters contain data that identifies a particular record
- Query params contain data that select customized behaviors
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+We may often see query params used for:
```
anselrognlie, line 104: a little less hedge-y
```suggestion
Common uses for query params include:
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+Within Flask, we can get the value from any query param using the `request.args` object.
+
+```python
+query_param_value = request.args.get(query_param_key)
+```
+
+| <div style="min-width:250px;"> Piece of Code </div> | Notes                                                                                  |
+| --------------------------------------------------- | -------------------------------------------------------------------------------------- |
+| `query_param_value =`                               | We can make a local variable to represent the value of the param                       |
```
anselrognlie, line 119: it doesn't represent the value. we store the value there
```suggestion
| `query_param_value =`                               | A local variable in which we store the looked up value of a query param for later use  |
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+| `request.args.get(...)`                             | Inside of a route, we can access the value of a query param using this method          |
+| `query_param_key`                                   | We should put in the query param key here, either as a variable or as a string literal |
+
+### !callout-warning
+
+## Pro-tip: Check `request` Spelling
+
+The popular package `requests`, used for making HTTP requests, is not the same as the `request` that represents the HTTP request.
```
anselrognlie, line 127: good idea to call this out, but I feel like providing a little more context about when to pay attention might help
```suggestion
The popular package `requests`, used for making HTTP requests, is not the same as the `request` that represents the HTTP request. If we are working on a web API project that also calls external APIs by using the `requests` package, we should be especially careful of whether we have written `request` or `requests`!
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+### !end-callout
+
+## Using Query Params
+
+How we use query params now depends on our project needs, and using whatever Python strategies we need!
```
anselrognlie, line 133: I don't think the use of query params has much to do with python intrinsically.
```suggestion
The way we choose to use query params will depend on our project needs.
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Finding Books by Title: Preparation
+
+Consider this feature:
+
+> As a client, I want to send a request trying to get a list of books with a matching title, so I know which books have a matching title.
+
+### Predict HTTP Requests, Responses, and Logic
```
anselrognlie, line 141: same change as elsewhere
```suggestion
### Planning HTTP Requests, Responses, and Logic
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+Let's consider the endpoint to get a single book:
```
anselrognlie, line 143: (seems like a copy paste error)
actively discuss why this modification goes with the existing get all route
```suggestion
We want to get a list of book results, so the base of our endpoint will look like our usual endpoint to get all records of a particular type. We will use the `GET` verb sent to the `/books` endpoint. But now we'd also like to provide an additional parameter `title` that we can use to filter the results. Since filtering is a customization of the default _get all_ behavior, we can decide to express that as a query param.
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+| HTTP Method | Endpoint              |
+| ----------- | --------------------- |
+| `GET`       | `/books?title=Apples` |
+
+No request body.
```
anselrognlie, line 149: acknowledge why we don't need it
```suggestion
As usual, we don't need a request body for `GET` requests.
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+This assumes that there is at least a `book` table with the following rows. (The contents of `id`, `description`, and `completed_at` are irrelevant for this feature).
```
anselrognlie, line 151: similar 'at least' change as elsewhere
completed at seems copy paste from task list
```suggestion
This assumes that there is a `book` table with at least the following rows. (The contents of `id`, and `description` are irrelevant for this feature).
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+| `title`   |
+| --------- |
+| `Apples`  |
+| `Oranges` |
+
+The response we want to send back is:
```
anselrognlie, line 158: remind what we want the response to do
```suggestion
In response to the request we've discussed, we want the endpoint to return success for the status code, and a list of the books with a matching title. We can use `200 OK` for the response status code, and a JSON response body representing a list of books.
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+| Response Status | Response Body                                            |
+| --------------- | -------------------------------------------------------- |
+| `200 OK`        | `[{"id": ..., "title": "Apples", "description": "..."}]` |
+
+During this method, we will need to:
```
anselrognlie, line 164: essentially the same transition as elsewhere
```suggestion
Now that we have planned out these changes to our endpoint behavior, we can turn our attention to how to implement them.

Our endpoint will need to:
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+1. Check if the HTTP method is `GET`
+1. Check if we have a query param for `title`
+1. If we have a `title` query param, retrieve all of the books from the db that match
```
anselrognlie, line 168: set up for a "pre-refactored" solution. If all we need to do is decide our source of books, we don't need to present a repetitive version where we have to generate the JSON book results twice. instead, we can jump to a better version where we get books from one of two branches, but the rendering remains unified.
```suggestion
1. If we have a `title` query param, retrieve all of the books from the database that match, otherwise retrieve all books as usual
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+1. Format the books data into the appropriate structure (list of dictionaries, where each dictionary has `id`, `title`, and `description`)
+1. Send back a response
+
+## Finding Books by Title: Pseudocode
```
anselrognlie, line 172: providing real code
```suggestion
## Finding Books by Title: Code
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+For this feature, we must return to our `GET` `"/books"` endpoint and re-work it.
```
anselrognlie, line 174: While bringing this into alignment with previous edits, I noticed that we're a little inconsistent around how we refer to the routes. We had been referring to the show-like route as /<book_id>, even though it should be /books/<book_id>. So I put in some wiggle room here for both /books and "", but I'm a little unhappy. I might just need to be okay with this...
```suggestion
This endpoint uses the same path as our existing `/books` route that lists books. Recall that the `"/books"` part comes from the blueprint, so our route path is set to `""`.
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+Let's look at this example pseudocode for one strategy:
```
anselrognlie, line 176: we're presenting actual code
```suggestion
Let's modify our endpoint code to filter the results when a title query param is supplied.
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+        # Check if the query param "title" exists
+        # by using request.args.get("title")
+            # Query the db for all matching books
+            # Create a response
+            # Return it
+        # Else if the query param "title" does not exist
+            # Query the db for all books...
+            # Create a response...
+            # Return it...
+    # ... no other changes
```
anselrognlie, line 191: revised with real code
went with handle_books for now. Change may be required with further PR?
The `if` that picks which `books` to use _could_ use ternary notation, but in the same way the examples avoid list comprehensions, I thought avoiding ternary would be better.
```suggestion
@books_bp.route("", methods=["GET", "POST"])
def handle_books():
    if request.method == "GET":
        # this code replaces the previous query all code
        title_query = request.args.get("title")
        if title_query:
            books = Book.query.filter_by(title=title_query)
        else:
            books = Book.query.all()
        # end of the new code

        books_response = []
        for book in books:
            books_response.append({
                "id": book.id,
                "title": book.title,
                "description": book.description
            })

        return jsonify(books_response)
    # ... existing code for creating a new book
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
```
anselrognlie, line 193: code description
```suggestion

| <div style="min-width:270px;"> Piece of Code </div> | Notes                                                                                                                                                           |
| --------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `title_query = ...`                                 | Store the result of looking for the `title` query param in the variable `title_query`                                                                           |
| `... = request.args.get("title")`                   | Try to get a query param called `title` from the `request`. This returns the value of the query param if it was set, or `None` if the query param is not found. |
| `if title_query:`                                   | Decide which conditional branch to take by checking whether we got a query param with which to filter                                                           |
| `books = ...`                                       | Store the results of our query in the `books` variable.                                                                                                         |
| `... = Book.query.filter_by( ... )`                 | If we got a query param, we will make a `filter_by` call to filter the results.                                                                                 |
| `title=title_query`                                 | Filter the book query results to those whose titles match the query param                                                                                       |
| `... = Book.query.all()`                            | If we didn't get a query param, get all the books as before.                                                                                                    |

The remainder of the code is the same as it was previously.

To summarize, we looked up whether the `title` query param was provided. We used that to decide which query we should run, and we stored the result in `books`. Then we converted `books` into JSON as before and returned the result.

```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+We can use Postman to manually test our database.
+
+We can compare our original feature which gets all books, to filtering by title.
+
+```
+GET localhost:5000/books
+GET localhost:5000/books?title=Apples
+```
+
+Our actual results will vary, depending on the contents of our databases. For example, if we've never created a `Book` with the title "Apple," then we'll never be able to retrieve that record!
```
anselrognlie, line 205: the "never able" makes it sound like we can never create a book with a title of Apple, so reworked that
```suggestion
Our actual results will vary, depending on the contents of our databases. For example, if we haven't created any `Book`s with the title "Apple," then the second test will always return an empty list!
```

more-flask-queries/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+
+Different records on different databases will change our tests and results.
```
anselrognlie, line 207: this sounds a bit throw-away. added the "predict, then test" to sneak in an additional takeaway
```suggestion
We should practice trying to predict the results of a test before running it to check our understanding. But we must take into account what records exist in our database, since this will affect our tests and results.
```

anselrognlie (changes requested):
comments for query params, finishing off this review pass

update-and-delete/update-and-delete.md
tildeee: Let's try it this cohort without the callout, and I'll look out for how that goes

more-flask-queries/query-params.md
tildeee: Just added these topic suggestions for Unit 3 lightning talks, thank you!!!

PR #9, opened 2021-03-30
Sorting algos

tildeee:


sorting-algos/sorting-algorithms.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Introduction
+
+Imagine a stack of index cards, where each index card has the numbers 1 through 100. Imagine that these cards are shuffled in a random, unknowable order.
+
+If you're given the task to sort these index cards in ascending order, what would you do?
+
+Picture yourself. Do you scan through the pile over and over again, slowly changing the order until it's all sorted? Do you use a table and make two separate piles? Do you compare the card you're looking at always to the next card in the pile?
+
+If you take two people and ask them to sort the same pile of cards, they will likely have different methods.
```
anselrognlie, line 30: "approaches" sounds more human to me than "methods" in this context
```suggestion
If you take two people and ask them to sort the same pile of cards, they will likely take different approaches.
```

anselrognlie (commented):
reviewed sorting-algorithms.md

continuing on, but there are a number of significant suggestions here, so I wanted to make the comments visible now.

sorting-algos/sorting-algorithms.md
```diff
@@ -0,0 +1,78 @@
...
+1. Define and describe the sorting strategy
+1. Visualize at least one example of this algorithm
+1. Learn the Big(O) complexity of this algorithm
+1. Practice reading the code of an example implementation
+
+Sorting is ordering a list of elements. We can distinguish two types of sorting. If the number of elements is small enough to fit in the main memory, sorting is called _internal sorting_. There are times when the number of elements is larger than can fit in the main memory, some of them will stay in the external storage while the sorting algorithm works and is called _external sorting_. We'll scope our conversation to a few internal sorting algorithms.
+
+## O(n<sup>2</sup>) Algorithms
+
+The sorting algorithms above have a time complexity of O(n<sup>2</sup>). These algorithms are fine when the array length, _n_, is small, but quickly cease to be useful as the size of the array increases. They **do** have the advantage of being relatively easy to implement and bubble sort and insertion sort can perform well in the best case, but they do not scale well. This brings in an overview of our first fairly performant sorting algorithm **Merge Sort**.
```
anselrognlie, line 62: This sounds like we have already discussed the n^2 algorithms and are getting ready to jump into merge sort (as it was in the textbook-curriculum).

Suggest the following change to be more of a forward-looking preview to reflect the new content ordering.

```suggestion
Several commonly introduced sorting algorithms have time complexities of O(n<sup>2</sup>). These include *bubble sort*, *selection sort*, and *insertion sort*, which will be discussed in detail in the following lessons.

These algorithms are fine when the array length _n_ is small, but quickly cease to be useful as the size of the array increases. Bubble sort and insertion sort can perform well in the best case, but they do not scale well.

They **do** have the advantage of being approachable for discussion.

Moving beyond these O(n<sup>2</sup>) algorithms, we will discuss *merge sort*, a O(n log n) algorithm with a beautiful symmetry!
```

anselrognlie (commented):
Forgot to include this suggestion in the sorting algorithms overview.

sorting-algos/bubble-sort.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Big(O) Complexity
+
+The time complexity of Bubble sort is _O(n<sup>2</sup>)_. We can conclude this by considering:
+
+- The inner most swap conditional statement is _O(1)_
+- The inner loop runs _i_ times i.e. _O(i)_ time complexity
+- The outer loop runs from _i_ = _0_ to _n_
+
+Overall, the inner loop runs _n-1_ times during the first iteration of the outer loop, _n-2_ times during the second iteration through the outer loop, and so on. i.e. _n-1 + n-2 + n-3 + ... + 3 + 2 + 1_ = _n(n-1)/2_ = _O(n<sup>2</sup>)_
```
anselrognlie, line 88: It's been a bit since the big O unit, so I'd like to expand this significantly (as the first of the 3 algorithms) to review concepts from the previous big O lesson and how they apply here.

```suggestion
There are two basic operations in which we are interested when analyzing sorting algorithms:

1. the number of comparisons
1. the number of swaps

In our big O analysis of sorting algorithms, we will typically focus on the number of comparisons. In some languages, the number of swaps will also be important and could provided a reason to favor an algorithm with fewer swaps over an algorithm with many swaps.

Also, recall that big O analysis considers the worst case for an algorithm! Although it's possible for bubble sort to finish after a single pass (_O(n)_) if the array happened to be sorted already, we are less interested in that situation.

Noticing best case performance can be useful if we are choosing between multiple algorithms that otherwise have the same complexity, or if we have forehand knowledge about the data with which we will be working.

In the worst case, we would need to perform the check and swap with each pair, and everything would be out of order so that we don't finish early. This would require about as many passes through the array as there are values.

This gives us a time complexity for bubble sort of _O(n<sup>2</sup>)_, where _n_ is the array length.

We conclude this by considering:

- The outer loop (overall number of passes) must run _n_ times
- The inner loop (performing comparisons and swapping) must also run _n_ times

We can even be a little more precise with our counting.

The inner loop runs _n-1_ times during the first iteration of the outer loop, _n-2_ times during the second iteration through the outer loop, and so on. i.e. _n-1 + n-2 + n-3 + ... + 3 + 2 + 1_.

There is a mathematical identity which states this summation is the same as _n(n-1)/2_. When multiplied out this results in _1⁄2n<sup>2</sup>-1⁄2n_, for _O(n<sup>2</sup>)_.

Notice that _n(n-1)/2_ does _not_ equal _n<sup>2</sup>_. Recall that when performing big O analysis, we are interested in the rate of growth in the algorithm rather than the exact number of operations. The growth rate will be dominated by the largest term in the equation.

After multiplying out the expression, the largest term is _1⁄2n<sup>2</sup>_. After dropping the constant, this allows us to say that bubble sort is _O(n<sup>2</sup>)_.
```

sorting-algos/bubble-sort.md
```diff
@@ -0,0 +1,151 @@
...
+
+### !callout-info
+
+## Mathematical Proofs Out of Scope
+
+The mathematical proof is out of scope for our discussion. However, there are great resources online that explain the math in-depth!
```
anselrognlie, line 94: some rewording and add FYC

```suggestion
In general, full mathematical proofs for the algorithms we discuss are out of scope for the curriculum. We will often apply analyses that are more intuitive than rigorous. However, there are great resources online that explain the math in-depth! Follow your curiosity!
```

sorting-algos/bubble-sort.md
```diff
@@ -0,0 +1,151 @@
...
+
+### !end-callout
+
```
anselrognlie, line 97: add stability discussion (as mentioned in the sorting lessons review)

```suggestion

## Stability

Bubble sort is a stable sorting algorithm. Since we only shift values by one position, and even then only if they are strictly out of order, equivalent values will never "jump" over one another.

That is, if the value 10 appeared in the list twice, the leftmost instance would always remain to the left, while the rightmost would remain to the right.

As discussed in the Sorting Algorithms lesson, stability can be a useful property for a sorting algorithm to have. If our values were more complex records that we would like to sort based on multiple attributes, then we can simply apply a stable sort multiple times. Or if the initial ordering of a list isn't entirely random, a stable sort will preserve the ordering between records that are otherwise equivalent.

```

sorting-algos/bubble-sort.md
```diff
@@ -0,0 +1,151 @@
...
+        while j < (len(array) - i - 1):
+            # Compare two adjacent items
+            if array[j] > array[j+1]:
+                # Swap
+                temp = array[j]
+                array[j] = array[j+1]
+                array[j + 1] = temp
+            j += 1
+        i += 1
+    return array
```
anselrognlie, line 118: consistent spacing for `[j+1]`
remove the return since this is modifying the passed in array (the suggested explanation below mentions this explicitly)

```suggestion
def bubble_sort(array):
    i = 0
    # Begin the outer loop
    while i < len(array) - 1:
        j = 0
        # Begin the inner loop
        while j < (len(array) - i - 1):
            # Compare two adjacent items
            if array[j] > array[j+1]:
                # Swap
                temp = array[j]
                array[j] = array[j+1]
                array[j+1] = temp
            j += 1
        i += 1
```

sorting-algos/bubble-sort.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+We should compare this code with our detailed explanation:
```
anselrognlie, line 121: consistent verbiage with the other lessons

```suggestion
Compare this code with this detailed explanation of the algorithm:
```

sorting-algos/bubble-sort.md
```diff
@@ -0,0 +1,151 @@
...
+
+- Start a variable `i` at `0`
+- Create an outer loop that loops while `i` is smaller than the length of the array - 1
+  - Start a variable `j` at `0`
+  - Create an inner loop that loops while `j` is smaller than `len(array) - i - 1`
+    - Compare the items at index `j` and `j+1`. If they're out of order:
+      - Swap the items at `array[j]` and `array[j+1]` using a temporary variable, `temp`
+    - Increment `j`
+  - Increment `i`
```
anselrognlie, line 130: expand the explanation, both "inline" as well as general observations

```suggestion
- Start a variable `i` at `0`
- Create an outer loop that loops while `i` is smaller than the length of the `array - 1` since we don't need to bubble the final value
  - Start a variable `j` at `0`
  - Create an inner loop that loops while `j` is smaller than `len(array) - i - 1` since we want to avoid the `i` sorted elements at the end, and we need to leave one extra value for the pairwise comparisons
    - Compare the items at index `j` and `j+1`. If they're out of order:
      - Swap the items at `array[j]` and `array[j+1]` using a temporary variable, `temp`
    - Increment `j`
  - Repeat until `j` reaches its limit
  - Increment `i`
- Repeat until `i` reaches its limit

This implementation does _not_ explicitly track whether a swap has been made (which could allow early termination). But when we get down to a single item in the unsorted portion, this is implicitly true, as there is nothing left with which we can compare.

This implementation sorts the array in-place. That is, it uses only a small amount of constant extra memory (just a few local variables), and the original array itself is updated. Because the original array is modified, no return statement is needed.

Note that the presented implementation used a temporary variable to perform the swap. This is a common pattern in languages that don't support multiple values in a single assignment. This implementation has been written in a general manner to emphasize that the algorithm—the steps for performing bubble sort—are not specific to any one language.

We will follow this same style when discussing other sorting algorithms. But we should feel free to modify the presented implementation to make better use of language-specific features!
```

anselrognlie (commented):
bubble sort feedback

sorting-algos/selection-sort.md
```diff
@@ -0,0 +1,97 @@
...
+1. Once you find it, you swap this smallest element found with the first element of the array.
+1. Then you look for the smallest element in the remaining array (the sub-array without the first element) and swap this element found with the second element.
+1. Then you look for the smallest element in the remaining array (the sub-array without the first and second elements) and swap that element with the third element, and so on.
+
+## Example
+
+Consider the initial unsorted array `[99, 45, 35, 40, 16, 50, 11, 7, 90]`. Here's what it looks like before and after each iteration of the outer loop.
+
+- `i` is the index that starts at 0, and increments until the length of the array.
+- `min_index` is a variable that holds the index of the smallest element in the sub-array.
```
anselrognlie, line 27: since we're showing an abbreviated table  (unlike the fully exploded bubble sort), I'd like to expand the description for min_index to be very explicit that we need an inner loop to find the min value.

```suggestion
- `min_index` is a variable that holds the index of the smallest element remaining in the unsorted portion of the list. To locate this item, we use an inner loop through the remaining unsorted values for each pass of the outer loop.
```

sorting-algos/selection-sort.md
```diff
@@ -0,0 +1,97 @@
...
+
```
anselrognlie, line 28: add guidance on how to interpret the following table

```suggestion

The array in each row is shown _before_ the swap for that pass is performed, so that the index under consideration `i` and the index of the minimum found value `min_index` line up with the displayed array values. The array shown on each following line has the previously identified swap applied.

```

sorting-algos/selection-sort.md
```diff
@@ -0,0 +1,97 @@
...
+| --------- | ------------------------------------------- | --- | ----------- |
+| 1.        | [**99**, 45, 35, 40, 16, 50, 11, **7**, 90] | 0   | 7           |
+| 2.        | [7, **45**, 35, 40, 16, 50, **11**, 99, 90] | 1   | 6           |
+| 3.        | [7, 11, **35**, 40, **16**, 50, 45, 99, 90] | 2   | 4           |
+| 4.        | [7, 11, 16, **40**, **35**, 50, 45, 99, 90] | 3   | 4           |
+| 5.        | [7, 11, 16, 35, **40**, 50, 45, 99, 90]     | 4   | 4           |
+| 6.        | [7, 11, 16, 35, 40, **50**, **45**, 99, 90] | 5   | 6           |
+| 7.        | [7, 11, 16, 35, 40, 45, **50**, 99, 90]     | 6   | 6           |
+| 8.        | [7, 11, 16, 35, 40, 45, 50, **99**, **90**] | 7   | 8           |
+| 9.        | [7, 11, 16, 35, 40, 45, 50, 90, 99]         | -   |             |
```
anselrognlie, line 39: similar table revisions as applied to bubble sort

```suggestion
| Pass | <div style="min-width: 300px;">Array</div>  | `i` | `min_index` |
| ---- | ------------------------------------------- | --- | ----------- |
| 1    | [**99**, 45, 35, 40, 16, 50, 11, **7**, 90] | 0   | 7           |
| 2    | [7, **45**, 35, 40, 16, 50, **11**, 99, 90] | 1   | 6           |
| 3    | [7, 11, **35**, 40, **16**, 50, 45, 99, 90] | 2   | 4           |
| 4    | [7, 11, 16, **40**, **35**, 50, 45, 99, 90] | 3   | 4           |
| 5    | [7, 11, 16, 35, **40**, 50, 45, 99, 90]     | 4   | 4           |
| 6    | [7, 11, 16, 35, 40, **50**, **45**, 99, 90] | 5   | 6           |
| 7    | [7, 11, 16, 35, 40, 45, **50**, 99, 90]     | 6   | 6           |
| 8    | [7, 11, 16, 35, 40, 45, 50, **99**, **90**] | 7   | 8           |
| 9    | [7, 11, 16, 35, 40, 45, 50, 90, 99]         | -   | -           |
```

sorting-algos/selection-sort.md
```diff
@@ -0,0 +1,97 @@
...
+
```
anselrognlie, line 40: add observation about final pass

```suggestion

Note that we do not actually run the 9th pass for this 9 value array. Since there is only one item left in the unsorted portion of the array and no available swap locations, we can end one pass "early."

```

sorting-algos/selection-sort.md
```diff
@@ -0,0 +1,97 @@
...
+## Big(O) Complexity
```
anselrognlie, line 41:
```suggestion
## Big O Complexity
```

sorting-algos/selection-sort.md
```diff
@@ -0,0 +1,97 @@
...
+
+The time complexity of Selection sort is _O(n<sup>2</sup>)_.
+
+Each of the nested loops has an upper bound defined by the count of elements, _n_ in the list to be sorted.
```
anselrognlie, line 45: I think it's worth still spending some time repeating the big O discussion to build comfort with the topic. It also includes some verbiage acknowledging other languages and how concerns that affect one language may not be the same as those in another language.

```suggestion
The time complexity of selection sort is _O(n<sup>2</sup>)_.

The way by which we reach this conclusion is very similar to our discussion of bubble sort.

Our outer loop runs about _n_ times. The first pass through, we must perform _n-1_ comparisons to find the minimum value, and then perform a single swap for that value.

Note this minor improvement over bubble sort, as bubble sort would perform both _n-1_ comparisons, as well as _n-1_ swaps. This does not affect our overall complexity analysis since the inner loop does still run _n-1_ times. But again, in some languages this could be important.

Each successive pass requires one fewer comparison, so we end up with the same decreasing summation as we did in bubble sort, which gives us a complexity of _O(n<sup>2</sup>)_.

For a more intuitive argument. Consider that both the outer loop (tracking the position to place the minimum) and the inner loop (looking for the current minimum) are related to the size of the array _n_, and since they are nested, we multiply their bounds together for _n*n_ or _O(n<sup>2</sup>)_.
```

sorting-algos/selection-sort.md
```diff
@@ -0,0 +1,97 @@
...
+
```
anselrognlie, line 46: include discussion of the sort stability

```suggestion

## Stability

Selection sort is _not_ a stable sorting algorithm. The swaps that we perform can result in equivalent values jumping over one another, losing the information about their relative initial orderings.

If we need a stable sorting algorithm, we should select a different algorithm, or think about how we could modify the base selection sort algorithm to enforce stability.

```

sorting-algos/selection-sort.md
```diff
@@ -0,0 +1,97 @@
...
+        while j < len(array):
+            if array[j] < array[min_index]:
+                min_index = j
+            j += 1
+        if min_index != i:
+            temp = array[min_index]
+            array[min_index] = array[i]
+            array[i] = temp
+        i += 1
+    return array
```
anselrognlie, line 66: remove the return since this sorts in input array in-place

```suggestion
```

sorting-algos/selection-sort.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+Compare this code with this detailed explanation of the algorithm:
+
+- Finds the smallest element in the range of `i` and `min_index`
+- Sets `min_index` to the index of the smallest item
+- If `i` and `min_index` are not the same, then
+  - Swap the elements at index `i` and index `min_index`
+- Increases `i`
+- Do this until `i` reaches the end of the list
```
anselrognlie, line 76: - bring the explanation up to the same level of detail as the bubble sort
- the mention of min_index in the first line seems off to me anyway
- include final observation about the sort occurring to the passed in array directly.

```suggestion
a variable `i` at `0`
- Create an outer loop that loops while `i` is smaller than the length of the `array - 1` since we don't need to swap the final value
  - Start with the assumption that the minimum value is already in the right place by setting `min_index` to `i` 
  - Start a variable `j` at the position one past `i` to look for the smallest element in the rest of the array
  - Create an inner loop that loops `j` over the remainder of the array
    - If the current inspected value is smaller than the current smallest value, update `min_index` with the current index `j`
    - Increase `j`
  - Repeat until `j` reaches the end of the list
  - If `i` and `min_index` are not the same, then
    - Swap the elements at index `i` and index `min_index` using a temporary variable `temp`
  - Increase `i`
- Repeat until `i` reaches the end of the list

This implementation sorts the array in-place. Because the original array is modified, no return statement is needed.
```

sorting-algos/sorting-algos.instructor.md
```diff
@@ -0,0 +1,13 @@
+# Instructor: Sorting Algos
+
+## Sorting Algorithms (Lesson)
+
+A huge reason why we're teaching sorting algorithms is because it's considered "common knowledge" among traditional CS education.
+
+I think it's worth mentioning that fact verbally, to help students understand what to get out of this.
+
```
anselrognlie, line 8: include note about how to make selection sort stable (left open in the reading)

```suggestion

## Selection Sort Stability

To make selection sort stable, instead of directly swapping the value at `min_index` with the value at `i`, we can all the intervening values up by one to open the spot where the selected value needs to go.  With this modification, selection sort closely resembles an in-place insertion sort.

```

anselrognlie (commented):
comments for selection sort

sorting-algos/insertion-sort.md
```diff
@@ -0,0 +1,113 @@
...
+| --------- | --------------------------------------- |
+| 1.        | [**99**, 45, 35, 40, 16, 50, 11, 7, 90] |
+| 2.        | [**45, 99**, 35, 40, 16, 50, 11, 7, 90] |
+| 3.        | [**35, 45, 99**, 40, 16, 50, 11, 7, 90] |
+| 4.        | [**35, 40, 45, 99**, 16, 50, 11, 7, 90] |
+| 5.        | [**16, 35, 40, 45, 99**, 50, 11, 7, 90] |
+| 6.        | [**16, 35, 40, 45, 50, 99**, 11, 7, 90] |
+| 7.        | [**11, 16, 35, 40, 45, 50, 99**, 7, 90] |
+| 8.        | [**7, 11, 16, 35, 40, 45, 50, 99**, 90] |
+| 9.        | [**7, 11, 16, 35, 40, 45, 50, 90, 99**] |
```
anselrognlie, line 48: similar changes as for previous algorithms

```suggestion
| Pass | Array                                   |
| ---- | --------------------------------------- |
| 1    | [**99**, 45, 35, 40, 16, 50, 11, 7, 90] |
| 2    | [**45, 99**, 35, 40, 16, 50, 11, 7, 90] |
| 3    | [**35, 45, 99**, 40, 16, 50, 11, 7, 90] |
| 4    | [**35, 40, 45, 99**, 16, 50, 11, 7, 90] |
| 5    | [**16, 35, 40, 45, 99**, 50, 11, 7, 90] |
| 6    | [**16, 35, 40, 45, 50, 99**, 11, 7, 90] |
| 7    | [**11, 16, 35, 40, 45, 50, 99**, 7, 90] |
| 8    | [**7, 11, 16, 35, 40, 45, 50, 99**, 90] |
| 9    | [**7, 11, 16, 35, 40, 45, 50, 90, 99**] |
```

sorting-algos/insertion-sort.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Big(O) Complexity
```
anselrognlie, line 50:
```suggestion
## Big O Complexity
```

sorting-algos/insertion-sort.md
```diff
@@ -0,0 +1,113 @@
...
+
+- 0 comparisons to insert the first element
+- 1 comparison to insert the second element
+- 2 comparisons to insert the third element
+- ... and so on
+- _n-1_ comparisons (worst case) to insert the last element
+
+Overall, this is _1 + 2 + 3 + ... + (n-1)_ = _O(n<sup>2</sup>)_
+
+The insertion sort has a complexity of O(n<sup>2</sup>).
```
anselrognlie, line 62: this text sounds like it's unaware that we have discussed much of this in other locations already (even in the original textbook)
add more connection to the rest of the material

```suggestion
As with the other sorting algorithms we have considered, with insertion sort we examine the number of comparisons performed in the inner loop.

The insertion sort algorithm requires (in the worst case):

- 0 comparisons to insert the first element
- 1 comparison to insert the second element
- 2 comparisons to insert the third element
- ... and so on
- _n-1_ comparisons to insert the last element

Overall, this is _1 + 2 + 3 + ... + (n-1)_, the same expression we saw for bubble sort and selection sort.

Therefore insertion sort has a complexity of _O(n<sup>2</sup>)_.
```

sorting-algos/insertion-sort.md
```diff
@@ -0,0 +1,113 @@
...
+
+Unlike selection sort, insertion sort has a best case time complexity of O(n) when the items are already sorted. In other words, insertion sort runs in nearly linear time on a nearly sorted list of elements.
```
anselrognlie, line 64: additional explanation about why the best case is O(n)
```suggestion
We might also notice that insertion sort, like bubble sort, has a best case time complexity of _O(n)_ if the list is already sorted. As we expand the sorted sub-part of the array, each new item is already in its correct spot, which we can determine with a single check per item. Since there are _n_ items, this gives a best case of _O(n)_.

Insertion sort generally improves in complexity the closer the list is to being sorted, that is, the more items are already relatively in their correct places. In other words, insertion sort runs in _nearly_ linear time on a _nearly_ sorted list of elements.
```

sorting-algos/insertion-sort.md
```diff
@@ -0,0 +1,113 @@
...
+
```
anselrognlie, line 65: add stability information

```suggestion

## Stability

Insertion sort is a stable sorting algorithm. Like bubble sort, we only move items in the list using adjacent swaps, and then only when they are strictly out of order.

```

sorting-algos/insertion-sort.md
```diff
@@ -0,0 +1,113 @@
...
+```python
+def insertion_sort(array):
+    i = 1
+    while i < len(array):
+        to_insert = array[i]
+        j = i
+        # Search in the sorted portion of the array
+        # for the correct position to insert array[i]
+        while j > 0 and array[j - 1] > to_insert:
+            array[j] = array[j - 1]
```
anselrognlie, line 79: spacing consistent with the previous algorithms
```suggestion
        while j > 0 and array[j-1] > to_insert:
            array[j] = array[j-1]
```

sorting-algos/insertion-sort.md
```diff
@@ -0,0 +1,113 @@
...
+            j -= 1
+        array[j] = to_insert
+        i += 1
+    return array
```
anselrognlie, line 83: no need for return since sorting in place

```suggestion
```

sorting-algos/insertion-sort.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+- Loop through the entire list
+- Find the first element of the unsorted list. This is the element that needs to be inserted, called `to_insert`
+- Find the correct index to insert the value `to_insert`
+  - Loop through the sorted list
+  - Compare the values between the item we're inspecting, and `to_insert`
+  - Swap if needed
+- Increase the indices, so that the sorted list grows, and the unsorted list shrinks
```
anselrognlie, line 92: include transition text used for other algorithms between example and explanation
slightly expand explanation (we're on the third one now, so a little less detail than the previous ones seems ok to me)
mention that no return is needed because in-place

```suggestion
Compare this code with this detailed explanation of the algorithm:

- Loop through the entire list with `i` representing the first position of the unsorted sub-array
- Store the first element of the unsorted list in a temporary variable called `to_insert`
- Find the correct index to insert the value `to_insert`
  - Loop through the sorted list from back to front
  - Compare the values between the item at the position we're inspecting, and `to_insert`
  - Swap if needed
- Increase the outer index `i` so that the sorted list grows, and the unsorted list shrinks

This implementation sorts the array in-place. Because the original array is modified, no return statement is needed.
```

anselrognlie (commented):
comments for insertion sort

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+## Learning Goals
+
+- Describe merge sort and its efficiency
+
+## Vocabulary and Synonyms
+
+| Vocab                        | Definition                                                                                                                                                         |
+| ---------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| Recursive Algorithm          | A method of solving a problem where the solution depends on solving smaller instances of the same problem. This approach can be applied to many types of problems. |
+| Divide-And-Conquer Algorithm | A strategy of solving a big problem by breaking it into smaller more managable subproblems. It can also lead to recursive solutions.                               |
```
anselrognlie, line 12: adjust column widths
minor punctuation

```suggestion
| <div style="min-width:200px;">Vocab</div> | Definition                                                                                                                                                         |
| ----------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| Recursive Algorithm                       | A method of solving a problem where the solution depends on solving smaller instances of the same problem. This approach can be applied to many types of problems. |
| Divide-And-Conquer Algorithm              | A strategy of solving a big problem by breaking it into smaller, more manageable subproblems. It can also lead to recursive solutions.                             |
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Merge Sort
```
anselrognlie, line 14:
```suggestion
## Overview
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+Merge sort is a _divide-and-conquer_ algorithm. It involves the following three stages:
+
+1. **Divide** the array into two sub-arrays at each step until each sub-array is of size one.
```
anselrognlie, line 18: "at each step" seems vague since we haven't discussed any steps

```suggestion
1. **Divide** the array into two sub-arrays repeatedly until each sub-array is of size one.
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+1. **Sort** each sub-array. (An array of size one is sorted by default.)
+1. **Merge** the sub-arrays into one array by combining two sub-arrays into one at each step.
+
+## Divide
+
+![Merge Sort Example](../assets/sorting-algos/MergeSort.png)
```
anselrognlie, line 24: replace image with divide-specific image

```suggestion
![Merge sort divide example. The list starts with the values (7, 2, 8, 1, 6, 5, 3, 9). This is split into to lists with one having (7, 2, 8, 1), and the other having (6, 5, 3, 9). Each list of four values is split into two lists of 2 value, resulting in 4 total lists. (7, 2), (8, 1), (6, 5), and (3, 9). Finally, each list of two is split into two single item lists for a total of 8 single item lists. (7), (2), (8), (1), (6), (5), (3), (9).](../assets/sorting-algos_merge-sort_divide.png)  
_Fig. Steps that merge sort takes when dividing a list containing 7, 2, 8, 1, 6, 5, 3, 9._
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+Let's reference this image to illustrate the divide step, starting at the top, original array.
+
+In the first _divide_ step, the original array of size eight gets divided into two sub-arrays of size four each. The arrays are divided at a calculated halfway point.
+
+In the next _divide_ step, we have two sub-arrays. Both of these arrays have four elements. They aren't of size one yet! So, the same action gets repeated. We calculate the halfway point, and divide the sub-arrays in half.
+
+This _divide_ stage continues until the original array of size _n_ is reduced to sub-arrays of size _1_ each.
+
+A sub-array of size one is trivially, and by default sorted.
```
anselrognlie, line 34: explain why 1-item arrays are trivially sorted. saying they're sorted "by default" implies there might be someway to make a single item array that _isn't_ sorted? But no, not possible.

```suggestion
An array of size one is trivially sorted. With only a single item in the list, there is no other item to be out of place!
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Sort and Merge
+
+The _merge_ stage starts by combining two sub-arrays at a time.
+
+While combining the sub-arrays containing `7` and `2` respectively, the values in each are compared and sorted. The merging process continues in this manner.
+
+### Sorting in Detail
+
+![Merge Sort Example](../assets/sorting-algos/MergeSort.png)
```
anselrognlie, line 44: replace with merge-specific diagram

```suggestion
![Merge sort merge example. We start with 8 individual lists. (7), (2), (8), (1), (6), (5), (3), (9). Each pair of single value lists is merged into a single list of two values, resulting in four lists of two values. (2, 7), (1, 8), (5, 6), (3, 9). Each pair of two value lists is merged into a single list of four values, resulting in two lists of four values. (1, 2, 7, 8), (3, 5, 6, 9). Finally the two lists of four items are merged into a single list of eight items. (1, 2, 3, 5, 6, 7, 8, 9).](../assets/sorting-algos_merge-sort_merge.png)  
_Fig. Steps that merge sort takes when merging the lists it previously divided._
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+| Comparing which two values? | Which is smaller and gets written to the auxiliary array? | Auxiliary array       |
+| --------------------------- | --------------------------------------------------------- | --------------------- |
+| 1, 3                        | 1                                                         | [1]                   |
+| 2, 3                        | 2                                                         | [1, 2]                |
+| 7, 3                        | 3                                                         | [1, 2, 3]             |
+| 7, 5                        | 5                                                         | [1, 2, 3, 5]          |
+| 7, 6                        | 6                                                         | [1, 2, 3, 5, 6]       |
+| 7, 9                        | 7                                                         | [1, 2, 3, 5, 6, 7]    |
+| 8, 9                        | 8                                                         | [1, 2, 3, 5, 6, 7, 8] |
```
anselrognlie, line 58: show the two lists that are being merged in the table (bold the values being compared)
column width adjustment

```suggestion
| <div style="min-width:100px">Comparing which values (in bold)?</div> | Which is smaller and gets written to the auxiliary array? | <div style="min-width:200px">Auxiliary array</div> |
| -------------------------------------------------------------------- | --------------------------------------------------------- | -------------------------------------------------- |
| [**1**, 2, 7, 8] <br /> [**3**, 5, 6, 9]                             | 1                                                         | [1]                                                |
| [1, **2**, 7, 8] <br /> [**3**, 5, 6, 9]                             | 2                                                         | [1, 2]                                             |
| [1, 2, **7**, 8] <br /> [**3**, 5, 6, 9]                             | 3                                                         | [1, 2, 3]                                          |
| [1, 2, **7**, 8] <br /> [3, **5**, 6, 9]                             | 5                                                         | [1, 2, 3, 5]                                       |
| [1, 2, **7**, 8] <br /> [3, 5, **6**, 9]                             | 6                                                         | [1, 2, 3, 5, 6]                                    |
| [1, 2, **7**, 8] <br /> [3, 5, 6, **9**]                             | 7                                                         | [1, 2, 3, 5, 6, 7]                                 |
| [1, 2, 7, **8**] <br /> [3, 5, 6, **9**]                             | 8                                                         | [1, 2, 3, 5, 6, 7, 8]                              |
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+At this final point, all elements of the first sub-array have been merged. All remaining elements of the second sub-array get copied over linearly to the auxiliary array.
+
+In this case, only one element is left in the second sub-array. So, _9_ gets copied over to the auxiliary array, for a final `[1, 2, 3, 5, 6, 7, 8, 9]`.
+
+This is the process of sorting and merging at the same time, which repeats until there are no more-sub arrays and the original array is completely sorted.
+
+Finally, the auxiliary array gets linearly copied back to the original array.
+
```
anselrognlie, line 67: add stability discussion

```suggestion

## Stability

Let's look at the stability of merge sort before tackling big O, since our complexity discussion will be a bit more involved.

Merge sort is a stable sorting algorithm, though we have to be a little careful to ensure this.

Recall that for a sorting algorithm to be stable, the sorted array must preserve the relative ordering between equivalent values that existed in the unsorted array. To ensure this, we need to guarantee that if there are two items with the same value, that the instance that started on the left, stays on the left.

For the stable _O(n<sup>2</sup>)_ algorithms, we did this bys ensuring that we never swapped "through" an equivalent value. But with all our dividing and merging, how can we ensure this remains true in merge sort?

Notice that when we divide a sub-array, there is always a left side and a right side. Anything that goes into the left array will originally have preceded anything in the right array. This is true all the way down during the divide phase. So if we encounter two equivalent values during the merge phase, one from the left and one from the right, we need only prefer the value from the left. As long as we do this all the way up through the merging, we will preserve the relative ordering required for a stable sort.

```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+## Big(O) Complexity
```
anselrognlie, line 68:
```suggestion
## Big O Complexity
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+The time complexity of merge sort is _O(n log n)_. Let's look closer to understand how we achieved this conclusion.
+
+### Divide Step
+
+Dividing and finding the midway index is a straightforward computation. This step takes constant time regardless of the sub-arrays.
```
anselrognlie, line 74: be a bit more precise with how the constant time split contributes to the overall complexity (overall it's a linear term)

```suggestion
Each divide step can be implemented as a straightforward computation to find the array midpoint from a start and end index. This step takes constant time regardless of the size of the sub-array, and there will be _n-1_ total divides. So the divide step would contribute _n-1_, a linear term, to the total complexity. We will be able to ignore this term since it will not dominate the complexity growth.
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Merge Step
+
+Merging a total of _n_ elements takes _O(n)_ time. If there are two sub-arrays of size _n/2_ each, then we will compare one element from one sub-array with another element from the second sub-array and one of the two will get copied. This step will continue until all are copied, taking a total of _O(n)_ time.
+
+As the sub-problems get smaller, the number of sub-problems doubles at each level, but the merging time halves. The doubling and halving cancel each other out and so the merging takes _O(n)_ time at each level of the merge steps (as seen in the image above).
```
anselrognlie, line 80: be a bit more specific about what to notice in the diagram

```suggestion
As the sub-problems get smaller, the number of sub-problems doubles at each level, but the merging time halves. The doubling and halving cancel each other out and so the merging takes _O(n)_ time at each level of the merge steps. Notice that each row in the merge diagram has _n_ (in this case 8) elements in total.
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Base Case
+
+In the base case, we have sub-arrays of size _1_ and a total of _n_ subarrays. It takes _O(1)_ time to sort an array of size one. Overall, merging at base level is _O(n)_ time, just like any other level.
```
anselrognlie, line 84:
```suggestion
In the base case, we have sub-arrays of size _1_ and a total of _n_ sub-arrays. It takes _O(1)_ time to sort an array of size one. Overall, merging the base level is _O(n)_ time, just like any other level.
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Count of levels
+
+Starting with _n_ elements and reducing by half at each level, until we reach one element sized array takes _log n_ steps. Similarly, starting with sub-arrays of one element each and combining two sub-arrays at a time until we reach an array of _n_ elements also takes _log n_ steps.
```
anselrognlie, line 88:
```suggestion
Starting with _n_ elements, and reducing by half at each level until we reach arrays with a single element each, takes _log n_ steps. Similarly, starting with sub-arrays of one element each and combining two sub-arrays at a time until we reach an array of _n_ elements also takes _log n_ steps.
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Conclusion
+
+Overall, each level takes _O(n)_ time. There are _O(log n)_ such levels.
+
+Therefore, O(n) times O(log n) levels results in an overall time complexity of _O(n log n)_.
```
anselrognlie, line 94: add italics to existing sentence
describe the overview of how we got this complexity result

```suggestion
Therefore, _O(n)_ merges times _O(log n)_ levels results in an overall time complexity of _O(n log n)_.

Notice that were able to arrive at this conclusion purely from a description of the operations that must be performed, and not from a direct inspection of any implementation.

In fact, when we encounter the implementation of merge sort later when discussing recursion, it might not be clear how to derive a big O analysis from it in the same way that we were able to for the _O(n<sup>2</sup>)_ sorting algorithms that used nested loops. In such cases, drawing a diagram of the operations can provide us with a geometric argument for thinking about the algorithm's complexity.
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+### !callout-info
+
+## Merge Sort Implementation Uses Recursion
+
+Understanding how to implement merge sort is best paired with the topic of recursion.
```
anselrognlie, line 100: expand into FYC

```suggestion
Understanding how to implement merge sort is best paired with the topic of recursion. We will revisit merge sort at that time, but if this sounds interesting, follow your curiosity!
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+### !end-callout
+
+### Performance Considerations
+
+For small arrays, insertion sort performs better than merge sort. However, as the size of the array grows, merge sort becomes much more performant.
```
anselrognlie, line 106: provide motivation for comparing insertion sort and merge sort (the current conclusion is moved to the end of this section)

```suggestion
It may seem like we have to do a lot of extra work for merge sort that we don't need to do for something like insertion sort. So how can merge sort have a time complexity of _O(n log n)_ while the simpler insertion sort has a time complexity of _O(n<sup>2</sup>)_?

It is absolutely accurate to say that insertion sort is a simpler algorithm with less bookkeeping overhead. But we should remember that big O is not a measure of absolute performance. It tells us how the performance grows with the size of the input.

This means that for small arrays, insertion sort may absolutely perform better than merge sort. However, as the size of the input grows, merge sort will tend to outperform insertion sort.
```

sorting-algos/merge-sort.md
```diff
@@ -0,0 +1,108 @@
...
+
+There are other O(n log n) sorting algorithms, notably [Quick Sort](https://www.geeksforgeeks.org/quick-sort/), and [Heap Sort](https://www.geeksforgeeks.org/heap-sort/).
```
anselrognlie, line 108: recast as a FYC callout and add a CFU take away

```suggestion
### !callout-info

## There Are Other _O(n log n)_ Sorting Algorithms

Other than merge sort, the most well known _O(n log n)_ sorting algorithms include [quick sort](https://www.geeksforgeeks.org/quick-sort/), and [heap sort](https://www.geeksforgeeks.org/heap-sort/). If this introduction to merge sort has been interesting, then diving into these other algorithms may also be rewarding. Follow your curiosity!

### !end-callout

## Check for Understanding

<!-- Question Takeaway -->
<!-- prettier-ignore-start -->
### !challenge
* type: paragraph
* id: d4d6b4de
* title: Merge Sort
##### !question

What was your biggest takeaway from this lesson? Feel free to answer in 1-2 sentences, draw a picture and describe it, or write a poem, an analogy, or a story.

##### !end-question
##### !placeholder

My biggest takeaway from this lesson is...

##### !end-placeholder
### !end-challenge
<!-- prettier-ignore-end -->
```

anselrognlie (commented):
merge sort comments

sorting-algos/problem-set-sorting-algorithms.md
```diff
@@ -0,0 +1,473 @@
...
+* id: rNnJPu
+* title: Sorting Algorithms
+##### !question
+
+Describe the algorithm for Bubble Sort in your own words. Make sure to touch on:
+
+- What makes this algorithm unique?
+- What is its time and space complexity?
+- What factors contribute to its time complexity?
+- What factors contribute to its space complexity?
```
anselrognlie, line 156: We don't really explicitly discuss the space complexity of the algorithms. For the n^2 ones, they are all in-place with O(1). But we kind of wave our hands for merge sort. It's still sort of presented as "in-place" but only by the use of a temporary list that gets copied back into the original list. A "true" in-place merge sort loses a lot of its efficiencies. (related: we should be careful about describing merge sort as in-place)

I'm ok with leaving this as something for the students to think about (of course the answers are a google away), so I'm assuming we're evaluating these responses more based on their individual thought process than on having a bullet list of "correct" answers?

sorting-algos/problem-set-sorting-algorithms.md
```diff
@@ -0,0 +1,473 @@
...
+* language: python3.6
+* id: P2sZ7P
+* title: Sorting Algorithms
+### !question
+
+There's nothing like getting familiar with sorting algorithms than to implement it yourself!
+
+Without using Python's `sort()` method (or anything like that), implement `sort_by_length` based on bubble sort, selection sort, or insertion sort.
+
+`sort_by_length` is a function that takes in one string. It returns a list of strings, where the items are the words from the string, ordered by length. The words with shorter length are placed before words with longer length.
```
anselrognlie, line 284:
```suggestion
`sort_by_length` is a function that takes in one string. It returns a list of strings, where the items are the words from the string, ordered by length. Words with shorter length are placed before words with longer length.
```

sorting-algos/problem-set-sorting-algorithms.md
```diff
@@ -0,0 +1,473 @@
...
+
+When words are tied for length, maintain the order they appeared in the original string.
```
anselrognlie, line 286: So this is a stable sort. Nice!

sorting-algos/problem-set-sorting-algorithms.md
```diff
@@ -0,0 +1,473 @@
...
+def test_sort_by_length_equal_length_maintains_order():
+    assert sort_by_length("words of equal length") == [
+        "of", "words", "equal", "length"]
+```
+
+### !end-question
+### !placeholder
+
+```python
+def sort_by_length(str):
```
anselrognlie, line 314: don't use `str` as a variable name
```suggestion
def sort_by_length(sentence):
    pass
```

sorting-algos/problem-set-sorting-algorithms.md
```diff
@@ -0,0 +1,473 @@
...
+```
+### !end-tests
+### !explanation
+
+An example of a working implementation:
+
+```python
+# Bubble Sort
+def sort_by_length(str):
+    array = str.split()
```
anselrognlie, line 347:
```suggestion
def sort_by_length(sentence):
    array = sentence.split()
```

sorting-algos/problem-set-sorting-algorithms.md
```diff
@@ -0,0 +1,473 @@
...
+
+The time complexity is... The space complexity is...
+
+##### !end-placeholder
+### !end-challenge
+<!-- prettier-ignore-end -->
+
+<!-- Question 11 -->
+<!-- prettier-ignore-start -->
+### !challenge
```
anselrognlie, line 386: what does this question have to do with sorting algorithms?

anselrognlie (changes requested):
reviewed activity (currently leaning towards skipping) and problem set.

commenting complete

config.yaml
```diff
@@ -0,0 +1,37 @@
+---
+Standards:
+  -
+    Title: Sorting Algos
```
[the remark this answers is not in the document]
tildeee: I'll absorb this in a separate PR... You know me, a stickler! And I want the topic title to match the folder name hahaha

sorting-algos/merge-sort.md
tildeee: This is awesome!!!!!!!

sorting-algos/problem-set-sorting-algorithms.md
tildeee: I think it'll be a good call to take this bullet point out, thanks for calling attention to it!

sorting-algos/problem-set-sorting-algorithms.md
tildeee: I LOVE this comment; I'm not convinced that this problem has a place in this problem set, either. I'm going to remove it!!!!!!! It'll help students focus on the previous question, the sorting one, better!!

PR #1, opened 2021-03-01
Data in Tables and Intro to SQL

tildeee:
https://docs.google.com/document/d/1ajzEx7wJt8dfU4md9YdaMpu7vrz0GSZ_Ch6kt3iJRCM/edit?usp=sharing

PR #10, opened 2021-03-31
Cm/more selecting in sql

CheezItMan:
Rough 1st pass with More Selecting in SQL


more-selecting-in-sql/grouping.md
```diff
@@ -0,0 +1,132 @@
...
+
+We could get a resultset back like the following:
+
+| state      | count |
+| ---------- | ----- |
+| Texas      | 10    |
+| Oregon     | 4     |
+| Washington | 230   |
+| New York   | 6     |
+| California | 15    |
```
anselrognlie, line 126: suggested replacement for this section (no suggestion because code)
as above, spend a bit more time developing the motivation for why certain columns can or cannot appear in a grouped select query.

(like any of these situation where a suggestion cannot be readily made, edit the comment to get the unaltered markdown.)

==============================

We saw in the previous example that listing a column (or the `*` wildcard) as part of the `SELECT` columns results in an error when they are not a part of the `GROUP BY` columns. Let's think about why this occurs.

SQL queries return tables of results. What kind of table could be the result of grouping? When we group by a particular column, all the records that share a particular value for that column are grouped together. Let's consider a subset of hypothetical data way back from cohort 8.

| first_name | cohort   |
| ---------- | -------- |
| Felicia    | Cohort 8 |
| Lorena     | Cohort 8 |
| Vera       | Cohort 8 |
| Agnes      | Cohort 8 |
| Courtney   | Cohort 8 |

_Fig. Sample data subset from the `alumni` table._

When we grouped by the cohort, we saw that this resulted in a unique list of column data. If we tried to include `first_name` in that result, how could a result table be generated? It would need to resemble the following:

```
 cohort   | first_name
----------+----------------------------------------
 Cohort 8 | Felicia, Lorena, Vera, Agnes, Courtney
(1 rows???)
```

There should be only one row for the cohort value, but many first names would need to be associated with that cohort value. This is not a valid result table structure, where each record lives in its own row.

Since we have included grouped results in the query, the only things we are allowed to include in the result are either the grouped columns themselves, or aggregate information about the grouped columns.

By using an aggregate function, such as `MAX`, `COUNT`, or any others we have discussed, we can get aggregate results for each grouped section of records.

The general syntax for a `SELECT` query using grouping is as follows:

```sql
SELECT
   column_desired_1,
   column_desired_2,
   ...,
   aggregate_function(column_to_aggregate)
FROM
   table_name
GROUP BY
   column_desired_1,
   column_desired_2,
   ...;
```

Notice that each of the entries in the `SELECT` portion are either columns corresponding to the columns listed in the `GROUP BY`, or they are an aggregate function applied to a non-grouped column.

For example, given this `books` table:

| title                         | price | author_id |
| ----------------------------- | ----- | --------- |
| Death of the Webbed Boa       | 10    | 1         |
| Groomsday                     | 5     | 1         |
| Beyond the Truth              | 6     | 2         |
| The Way of the Eight          | 12    | 3         |
| Mask of Exile                 | 11    | 2         |
| 2938 Destiny                  | 2     | 2         |
| Mystery of the Crispy Tourist | 16    | 3         |
| Clue of the Painted Pendant   | 25    | 3         |

_Fig. Example `books` table_

We could write the following query:

```sql
SELECT author_id, MAX(price)
FROM BOOKS
GROUP BY author_id
ORDER BY author_id;
```

This query will find the most expensive price for each author.

```
 author_id | max
-----------+-----
         1 | 10
         2 | 11
         3 | 25
(3 rows)
```

We could also write a query to find the number of alumni at Ada by state.

If our `alumni` table had the same schema as above (reproduced here):

Alumni Table  
_Fig. Ada `alumni` table reproduced from above_

We could find the number of alumni from each state with:

```sql
SELECT state, count(*)
FROM alumni
GROUP BY state;
```

The results might resemble the following:

```
 state      | count
------------+-------
 Texas      | 10
 Oregon     | 4
 Washington | 230
 New York   | 6
 California | 15
(5 rows)
```

more-selecting-in-sql/grouping.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Check for Understanding
+
+<!-- Question on SELECTING distinct Column entries -->
+
+<!-- SQL Question to SELECT the most recent birthdate by cohort -->
```
anselrognlie, line 132: adds takeaway. suggest other questions wait for a follow on pass

```suggestion
<!-- SQL Question to SELECT the most recent birthdate by cohort -->

<!-- Question Takeaway -->
<!-- prettier-ignore-start -->
### !challenge
* type: paragraph
* id: 05d33b23
* title: Grouping Results
##### !question

What was your biggest takeaway from this lesson? Feel free to answer in 1-2 sentences, draw a picture and describe it, or write a poem, an analogy, or a story.

##### !end-question
##### !placeholder

My biggest takeaway from this lesson is...

##### !end-placeholder
### !end-challenge
<!-- prettier-ignore-end -->
```

anselrognlie (changes requested):
Got through the files related to more selecting in sql. I expect @tildeee will need to review further for merging. (is there a way to reassign the owner/contact for a PR???)

We should probably sit on this until PR #1 is merged in, and then we can merge main here and hopefully a lot of noise will drop out. 🤞

more-selecting-in-sql/compound-conditions.md
```diff
@@ -0,0 +1,234 @@
...
+
+Like in other languages, these three operators have an order of precedence. For Postgres, this is:
+
+- `NOT`
+- `AND`
+- `OR`
+
+Rather than relying on the precedence ordering, we can also be explicit about how we want our condition to be evaluated by using parentheses.
+
+Let's try getting all the Intrigue genre books by either Madie McClure or Tim McDermott. We can start with either of our previous queries which returned the books by Madie McClure or Tim McDermott, and make a small modification to add the extra genre restriction.
```
tildeee, line 168: I crave a header to separate that the above paragraph is all the new info, and below the header is demonstrating it

```suggestion
#### Example

Let's try getting all the Intrigue genre books by either Madie McClure or Tim McDermott. We can start with either of our previous queries which returned the books by Madie McClure or Tim McDermott, and make a small modification to add the extra genre restriction.
```

PR #11, opened 2021-04-02
Testing our apis

tildeee:


testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
+# Environments and Env Vars
+
+## Goals
+
+Before we get to automated testing for our API, we'll need to learn about managing different environments.
+
+Our goal is to:
+
+1. Define what are development and testing environments
+1. Define environment variables and manage them
```
anselrognlie, line 10: revise the goals (plural, since more than one)
we don't go much into what "managing" means so I'd remove that from the second goal

```suggestion
Our goals are:

1. Define development and testing environments
1. Define environment variables
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+
```
anselrognlie, line 11: if we still want to call out "manage" consider adding a dedicated third goal about managing
```suggestion
1. Start thinking about how to manage environment variables

```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+## Environments
+
+In software, the word "environment" is used in many different ways. The meaning of "environment" changes based on the context.
+
+### !callout-secondary
+
+## Realistic Expectations
+
+No, really. "Environment" is a common concept among many different situations. When in doubt, clarify with questions!
```
anselrognlie, line 30: saying it's a common concept sounds like it should have the same meaning everywhere
rephrase to emphasize the definition is overloaded
```suggestion
No, really. "Environment" can mean many different things in many different situations. When in doubt, clarify with questions!
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+
+### !end-callout
+
+In the context of local project development and testing, there are two common patterns:
```
anselrognlie, line 34: this is showing a single pattern with two parts
```suggestion
In the context of local project development and testing, we will commonly use two environments according to the following pattern:
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+
+1. A _development environment_ dedicated for rapid development
+1. A _testing environment_ dedicated for holding test data
+
+| Name                    | Purpose                                                                                        | Differences                          |
```
anselrognlie, line 39: Differences sounds a little vague given the actual values in the table
```suggestion
| Name                    | Purpose                                                                                        | <div style="min-width:200px">Configuration Details</div>                          |
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+| ----------------------- | ---------------------------------------------------------------------------------------------- | ------------------------------------ |
+| Development environment | Dedicated to rapid development. Things will change often as the developer works on the project | Has a dedicated development database |
```
anselrognlie, line 41: explain what's in the development database, and call out debugging/logging
```suggestion
| Development environment | Dedicated to rapid development. Things will change often as we work on the project | Has a dedicated development database where we can add sample data we will use during development. May enable a debugging mode with detailed logging and error reports. |
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+| Testing environment     | Dedicated to setting up the data for running test suites.                                      | Has a dedicated testing database     |
```
anselrognlie, line 42: what's special about the testing database?
```suggestion
| Testing environment     | Dedicated to running test suites.                                      | Has a dedicated testing database containing specific data used to exercise our test suite.     |
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+## Realistic Expectations
+
+No, really. "Environment" is a common concept among many different situations. When in doubt, clarify with questions!
+
+### !end-callout
+
+In the context of local project development and testing, there are two common patterns:
+
+1. A _development environment_ dedicated for rapid development
+1. A _testing environment_ dedicated for holding test data
```
anselrognlie, line 37: 'to' sounds a little better to me and has agreement with the phrasing in the table
```suggestion
1. A _development environment_ dedicated to rapid development
1. A _testing environment_ dedicated to holding test data
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+
+| Name                    | Purpose                                                                                        | Differences                          |
+| ----------------------- | ---------------------------------------------------------------------------------------------- | ------------------------------------ |
+| Development environment | Dedicated to rapid development. Things will change often as the developer works on the project | Has a dedicated development database |
+| Testing environment     | Dedicated to setting up the data for running test suites.                                      | Has a dedicated testing database     |
+
+As developers, we'll have access to both environments. Whenever we run our Flask API, we can designate it to use a specific environment.
```
anselrognlie, line 44: we don't just _have_ access, we access them both a lot.
our discussions so far have made it sound like dev and test are completely distinct (and they can be), but since we'll be doing dev and test from our local machine, provide some explanation of why that is.
```suggestion
As developers, we access both environments.

As we develop our Flask API, we will work in our development environment. As we make changes to our code, we will test it frequently using our test environment. 

These two environments could exist on different machines. In practice, it is usually more convenient to have a way to run under either environment from our own developer machine. This allows us to test our code as we are working on it, minimizing the chance of breaking previous functionality (called a _regression_).

Whenever we run our Flask API, we can designate which environment to use.
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+
+The environment that our Flask API uses will determine which database it accesses while it runs.
```
anselrognlie, line 46: the environment can affect even more, so keep this at the level of general customization, but provide the database as an example of this
```suggestion
We will customize how we set certain configuration settings in our Flask API—such as the database it will access—based on the environment.
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+
+When Aleesha dances, she picks an environment. Different environments are configured to use different resources.
+
+Each environment has lighting. However, the lighting equipment doesn't, and shouldn't, be identical between the different environments:
+
+- Her practice setup at home has some small LED lights
+- The team's practice studio has adjustable ceiling lights
+- The main performance stage has professional lighting equipment
+
+Keeping these environments separate can help her have better expectations for what's available.
```
anselrognlie, line 58: having expectations about what's available sounds like we're bracing ourselves for disappointment. try focussing on the clarity and focus we gain by using different environments for different purposes.
```suggestion
Keeping these environments separate helps Aleesha and the other dancers focus on doing specific things in each environment.
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+
+We can imagine Aleesha's practice room at home is like a _development environment_.
```
anselrognlie, line 60: like I wrote above, we are basically using both dev and test together, so tie these concepts back to the metaphor.
```suggestion
We can imagine Aleesha's practice room at home is like her combined _development_ and _test_ environment.

As her _development_ environment, she can develop new dance routines in her own space. She can easily reconfigure the lighting and arrangement of things in that space to suit her work.

As her _test_ environment, she can practice and practice and make sure she has everything down before meeting with the rest of her dance team.
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+
```
anselrognlie, line 61: Our metaphor has the studio and main stage parts as well. I'd like to explain where they fit into dev practice as well.
```suggestion

### !callout-info

## More Environments Are Possible

For now, we will focus on development and test environments. But there are other environments that are also useful.

<br />

We can have _staging_ environments, which are more like the practice studio used by Aleesha's whole team. It's a place where all the work they've done individually comes together and can be seen working as a whole. Since this is a shared space involving multiple people, it's harder to make changes here.

<br />

We will also have a _production_ environment. This is the main performance stage, where the rest of the world sees our hard work. Once the performance starts, we _really_ don't want to make changes!

### !end-callout

```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+### Applying Environments at a High-Level
+
+As we introduce testing into our Flask API development, we will want to setup a dedicated testing environment. This will make our testing experience more consistent and reliable.
```
anselrognlie, line 64: explain why we want this and how it makes the experience better
```suggestion
As we introduce testing into our Flask API development, we want it to be easy for our tests to run using our test environment settings.

Having dedicated test settings is important so that our tests can run under the actual conditions we want to test. Making this change automatic will help us avoid running tests using the wrong data.

This will make our testing experience more consistent and reliable.
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Environment Variables
+
+Environment variables are strings that hold data outside of our application code. These variables are set at the environment level, and affect how an app is run.
```
anselrognlie, line 68: the jump from environments to variables feels disjoint, so included a small transition sentence.
as a reader _ don't know that I can identify with something being set  "at the environment level". I modified it to say that our environments will provide ways to set this (there's an added callout a bit later showing a small example)
give some overview of the kinds of affects they have on our code. (the details come in later bullets)

```suggestion
We will use environment variables to store information about values that change between environments.

Environment variables are strings that hold data outside of our application code. The environments we use provide ways to set these kinds of values.

Within our code, we can examine these variables, and either use them directly as data, or use them to make decisions about how our code should behave.
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+
+Some examples include:
+
+- Where the location of a database is
+- Where the location of other resources are (such as other services part of a larger software system)
```
anselrognlie, line 73: don't need where (especially compared to the other bullets)
callback to connection string vocab
I kept reading part as a noun which was throwing my ability to parse the sentence
```suggestion
- The location of a database (a connection string)
- Locations of other resources (such as other services that are part of a larger software system)
```

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+- Various configurations about how to run the app, such as debugging mode
+- API keys
+
```
anselrognlie, line 76: Add a callout showing a simple example of seeing and modifying environment variables
no suggestion because code
following comment includes the markdown

anselrognlie: ### !callout-info

## There Are Many Environment Variables Set Already

From any terminal prompt, we can run the command `env`. This displays the environment variables currently set in our terminal. We can set our own environment variables from the terminal as well. Let's run the following:

```bash
$ export MY_NEW_VAR="my value"
```

<br />

If we then run `env` again, we will see our environment variable `MY_NEW_VAR` appear among the output.

### !end-callout

testing-our-apis/environments-and-env-vars.md
```diff
@@ -0,0 +1,90 @@
...
+Additionally, environment variables often hold _sensitive and secret data_.
+
+In both of these cases, we will need to manage them. Managing environment variables lets us:
+
+- Keep them organized together, rather than as constant variables throughout the project code
+- Control their visibility
+
+It's often the case that we _must_ keep environment variables secret to the public. Because our codebases are often online, it may be necessary to **not commit environment variables to source control**.
+
+Different tools, packages, and practices will give us ways to manage and hide our environment variables.
```
anselrognlie, line 90: wound reorganizing this a bit
maintain the emphasis on avoiding committing
built a bit of a story about how we got from "values outside our code" to something we might accidentally commit.
included a callout emphasizing that we will be looking at just a small part of what goes into configuring environments.

```suggestion
Because environment variables are held outside of our code, they do not get checked into source control. This makes them very useful for storing _sensitive and secret data_.

- We might need a username and password to connect to a database.
- An API key is usually unique to our account. If someone else gets it, they can misuse a service acting under our identity.

As our applications grow, we may need to store a larger number of variables, to the point that using the `export` command becomes a problem. It would be nice if we could group our environment variables into a special file: a file that holds only our environment variables, and no code.

But once we have placed them in a file, we might accidentally commit that file. Because of the danger of misuse, we want to **avoid committing environment variables to source control**, where nefarious individuals could steal our credentials!

It seems like we're right back where we started!

In the next lesson, will explore how to manage our environment variables so that they remain organized, out of our code, and secure!

### !callout-info

## Many Ways to Customize Environments

Flask doesn't require a particular approach to setting environment configurations. We will keep our discussion of configuration focused on switching between databases and storing application secrets. Customizing configuration is a much larger topic, with many clever approaches to explore. Follow your curiosity!

### !end-callout
```

anselrognlie (commented):
comments for environements-and-env-vars.md

this felt a little murky about what an environment was (is it our terminal, is it different machines, is it a database). Flask also has the FLASK_ENV variable that we don't discuss. I leaned into dance metaphor, but split it up a bit more according to my understanding. hopefully this jives with your objectives!

testing-our-apis/dot-env-and-the-test-db.md
```diff
@@ -0,0 +1,193 @@
...
+  title VARCHAR(200),
+  description TEXT
+);
+```
+
+## `create_app` Now Handles Test Mode
+
+Now that we've set up environment variables and the test database, let's change some configurations in our app to use this.
+
+Based on the current state of this Flask app, let's look at a replacement for the `app/__init__.py` file, then go through the differences.
```
anselrognlie, line 124: minor revision in first passage
I'd just say this is revised and call out what to watch for
minor changes to discussion reminder
```suggestion
Now that we've set up environment variables and the test database, let's change some configurations in our app to make use of them.

Let's look at a revised version of the `app/__init__.py` file that includes the changes needed to handle test mode. We'll examine the differences following the code.
```

testing-our-apis/dot-env-and-the-test-db.md
```diff
@@ -0,0 +1,193 @@
...
+import os
+
+db = SQLAlchemy()
+migrate = Migrate()
+load_dotenv()
+
+def create_app(test_config=None):
+    app = Flask(__name__)
+
+    if test_config is None:
```
anselrognlie, line 140: I assume we'll want to make a change like this following the slack discussion?
```suggestion
    if not test_config:
```

testing-our-apis/dot-env-and-the-test-db.md
```diff
@@ -0,0 +1,193 @@
...
+        app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
+        app.config["SQLALCHEMY_DATABASE_URI"] = os.environ.get(
+            "SQLALCHEMY_TEST_DATABASE_URI")
+
+    db.init_app(app)
+    migrate.init_app(app, db)
+
+    from app.models.book import Book
+    with app.app_context():
+        db.create_all()
```
anselrognlie, line 155: this will get done in the test runner code
```suggestion
```

testing-our-apis/dot-env-and-the-test-db.md
```diff
@@ -0,0 +1,193 @@
...
+
+    from .routes import books_bp
+    app.register_blueprint(books_bp)
+
+    return app
+```
+
+| <div style="min-width:250px;"> Piece of Code </div> | Notes                                                                                                                                                                                                                                                              |
```
anselrognlie, line 163: improve the word wrapping
```suggestion
| <div style="min-width:310px;"> Piece of Code </div> | Notes                                                                                                                                                                                                                                                              |
```

testing-our-apis/dot-env-and-the-test-db.md
```diff
@@ -0,0 +1,193 @@
...
+| --------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `from dotenv import load_dotenv`                    | The `python-dotenv` package specifies to import the package like this                                                                                                                                                                                              |
+| `import os`                                         | This built-in package will allow us to read the environment variables set in `.env`                                                                                                                                                                                |
```
anselrognlie, line 166: this is a module, not a package
I'd just mention that it reads environment variables in general (even those not set through .env)
```suggestion
| `import os`                                         | This built-in module provides a way to read environment variables                                                                                                                                                                                |
```

testing-our-apis/dot-env-and-the-test-db.md
```diff
@@ -0,0 +1,193 @@
...
+| `load_dotenv()`                                     | The `python-dotenv` package specifies to call this method and load the variables before using them                                                                                                                                                                 |
```
anselrognlie, line 167: here we can mention that this bridges between the .env file and the os module
add explanations for why we are adding the param to create_app
```suggestion
| `load_dotenv()`                                     | The `python-dotenv` package specifies to call this method, which loads the values from our `.env` file so that the `os` module is able to see them.                                                                                                                                                                  |
| `def create_app( ... ):` | Add a new keyword parameter to `create_app` that we can use to customize its behavior |
| `test_config=None` | We have called the new parameter `test_config` which should receive a dictionary of configuration settings. It has a default value of `None`, making the parameter optional. |
```

testing-our-apis/dot-env-and-the-test-db.md
```diff
@@ -0,0 +1,193 @@
...
+| `if test_config is None:`                           | Create a conditional based on the keyword argument `test_config`. Whenever we call `create_app()`, from now on, we can optionally pass in a test config dictionary. If `test_config` is `None`, that means we are not trying to run the app in a test environment. |
```
anselrognlie, line 168: update for code agreement
most of this is moved up to the suggested new descriptions of the change to create_app
```suggestion
| `if not test_config:`                           | Check the keyword argument `test_config`. When we call `create_app()`, if `test_config` is falsy (`None` or empty), that means we are not trying to run the app in a test environment. |
```

testing-our-apis/dot-env-and-the-test-db.md
```diff
@@ -0,0 +1,193 @@
...
+| `os.environ.get( ... )`                             | This syntax gets an environment variable by the passed-in name                                                                                                                                                                                                     |
+| `"SQLALCHEMY_TEST_DATABASE_URI"`                    | This is the exact name of the **testing** database environment variable we defined in `.env`                                                                                                                                                                       |
+
+### !callout-secondary
+
+## Practical Tip: Reading > Writing
+
+Understanding how to read this code is more valuable than learning how to write it. Realistically, on many projects, test environments are set up. However, it's valuable to read this code in order to:
```
anselrognlie, line 180: untangle the line about setting up test environments: test environments get set up , and???
```suggestion
Understanding how to read this code is more valuable than learning how to write it. On many projects, we set up test environments by referring back to code from a previous project. But we should strive to be able to read this code in order to:
```

testing-our-apis/dot-env-and-the-test-db.md
```diff
@@ -0,0 +1,193 @@
...
+
+### !end-callout
+
+## Verify We Didn't Break It: Manually Test Development
+
+We'll test our test environment after we set up our ability to run tests!
+
+However, let's verify that we didn't break our development environment. Our development environment should not have changed at all!
+
+Let's run our app in development mode with `(venv) $ flask run`, and make some requests in Postman. We can manually test all of our functionality to ensure we haven't lost anything.
```
anselrognlie, line 193: some minor changes to the first 2 passages
slipped in a reference to regressions (introduced in a suggestion in the previous lesson)
```suggestion
We'll test our test environment after we set up tests in the next lessons!

For now, let's verify that we didn't break our development environment. These changes should not have affected our development environment at all!

Let's run our app in development mode with `(venv) $ flask run`, and make some requests in Postman. We can manually test all of our functionality to ensure we haven't introduced any _regressions_, broken scenarios which used to work.
```

anselrognlie (commented):
comments for dot-env-and-the-test-db

testing-our-apis/fixtures.md
```diff
@@ -0,0 +1,324 @@
+# Fixtures
```
anselrognlie, line 1: Updated fixtures.md with content

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
+# Test Setup
+
+## Introduction
+
+We can continue to use tests and pytest to test our Flask APIs! Reading, writing, and using pytest for a Flask API isn't very different from using pytest outside a Flask API.
```
anselrognlie, line 5: minor rewording to flow a bit better from the Fixture lesson. We've discussed a lot of this functionality in the previous lesson, so it should be less surprising that we'll be using tests in general, and pytest specifically.
```suggestion
We can build on our pytest knowledge to test our Flask APIs! Reading, writing, and using pytest for a Flask API isn't very different from using pytest outside a Flask API.
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+
+One main difference is the test setup. In this lesson, we'll configure our Flask projects to use pytest and a test database.
```
anselrognlie, line 7: specify that this _is_ the main difference. one main difference leaves the door open for the reader to worry about what other differences are going to pop up.
```suggestion
The main difference is in how we set up our tests. In this lesson, we'll configure our Flask projects to use pytest and a test database.
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Hello Books API
+
+This lesson uses the [Hello Books API](https://github.com/AdaGold/hello-books-api).
+
```
anselrognlie, line 12: minor formatting
```suggestion

<br />

```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+<details style="max-width: 700px; margin: auto;">
+    <summary>
+        Before beginning this lesson, the Hello Books API should have the following.
+    </summary>
+
+- A `hello_books_development` database
+- A `hello_books_test` database
+- A `book` table defined in both databases
```
anselrognlie, line 20: We haven't made (and won't explicitly make) the books table in the test database. We will be bypassing alembic's migration tracking and use the create_all/drop_all directly to set up and tear down the database with every run.
```suggestion
- A `book` table defined in the development database
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+- A `Book` model defined
+
+An `.env` file that contains:
```
anselrognlie, line 23: I read `.env` as dotenv, so we can use a instead of an.
```suggestion
A `.env` file that contains:
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+
+```
+SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://postgres:postgres@localhost:5432/hello_books_development
+SQLALCHEMY_TEST_DATABASE_URI=postgresql+psycopg2://postgres:postgres@localhost:5432/hello_books_test
+```
+
+Endpoints defined for these RESTful routes. They handle missing books:
```
anselrognlie, line 30: same change as in previous lesson
```suggestion
Endpoints defined for these RESTful routes. They can gracefully handle missing books:
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+- `tests/test_routes.py`
+
+We can do so with the following commands:
+
+```
+(venv) $ mkdir tests
+(venv) $ touch tests/__init__.py tests/conftest.py tests/test_routes.py
+```
+
+Let's consider what these files will do:
```
anselrognlie, line 77: consider sounds more like we have some familiarity and are thinking more deeply now, but we are getting our first look here.
```suggestion
Let's take a look at what these files will do:
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+
+| <div style="min-width:250px;">File</div> | Responsibility of this file                                                                                                                   |
+| ---------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
+| `__init__.py`                            | Establishes our `tests` package, so it can be properly connected with the rest of the app folders and files. We will probably leave it empty. |
```
anselrognlie, line 81: cut probably
```suggestion
| `__init__.py`                            | Establishes our `tests` package, so it can be properly connected with the rest of the app folders and files. We will leave it empty, as we typically do. |
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+| `conftest.py`                            | A conventional Flask file that will hold test configurations and common test helper functions                                                 |
```
anselrognlie, line 82: conftest.py isn't flask specific. It's a standard file in pytest suites.
```suggestion
| `conftest.py`                            | A standard pytest file that holds test configurations and common test helper functions. Essentially, it is automatically `import`ed before any other test files.                                                  |
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+from app import create_app
+from app import db
+
+
+@pytest.fixture
+def app():
+    app = create_app({"TESTING": True})
+
+    with app.app_context():
+        yield app
```
anselrognlie, line 100: add line to create the database tables (moved from the app init)
```suggestion
        db.create_all()
        yield app
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+| --------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `import pytest`                                     | We'll be utilizing pytest fixtures to set up our test app                                                                                                                                                   |
+| `from app import ... db`                            | We should import `create_app` and `db` in order to configure those when running the tests                                                                                                                   |
+| `@pytest.fixture`                                   | We'll create and use a pytest fixture named `app`, which will be used in our `client` fixtured (defined below)                                                                                              |
```
anselrognlie, line 115: fixtured → fixture
the fixture is defined below the code we are discussing, but defined about the discussion itself. defined later sounds a little more "relative" to me.
```suggestion
| `@pytest.fixture`                                   | We'll create and use a pytest fixture named `app`, which will be used in our `client` fixture (defined later)                                                                                              |
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+| `app = create_app( ... )`                           | When we run our tests, this line will run and create an `app` object. It's using the same `create_app` function defined in our `app/__init__.py` file!                                                      |
+| `{"TESTING": True}`                                 | Here, we're passing in a dictionary to represent a "test config" object. If we check the current implementation of `create_app()` in `app/__init__.py`, it uses this argument only to check if it's truthy. |
```
anselrognlie, line 117: don't need the "if we check"
```suggestion
| `{"TESTING": True}`                                 | Here, we're passing in a dictionary to represent a "test config" object. The current implementation of `create_app()` in `app/__init__.py` uses this argument only to check if it's truthy (which this is!). |
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+| `with app.app_context():`                           | This syntax designates that the following code should have an _application context_, or in this case, have access to objects such as `db`.                                                                  |
```
anselrognlie, line 118: Give a bit of a hand-wavey overview of what this does, but still tie it back to the database access that we're doing here.
```suggestion
| `with app.app_context():`                           | This syntax designates that the following code should have an _application context_. This lets various functionality in Flask determine what the current running app is. This is particularly important when accessing the database associated with the app.                                                                   |
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+| `yield app`                                         | This is pytest fixture syntax which executes the actual test at this moment. `yield app` finishes executing once the test has actually be run.                                                              |
```
anselrognlie, line 119: `yield` has now been discussed in the fixtures lesson, so we can be a little more precise in the description
```suggestion
| `yield app`                                         | This fixture suspends here, returning the app for use in tests or other fixtures. The lines after this `yield` statement will run after the test using the app has been completed.                                                              |
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+| `with app....: db.drop_all()`                       | After the test runs, this code specifies that we should clear our test database from any data that was saved inside our test.                                                                               |
```
anselrognlie, line 120:
```suggestion
| `with app....: db.drop_all()`                       | After the test runs, this code specifies that we should clear our test database of any data that was saved inside our test.                                                                               |
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+| `@pytest.fixture`                                   | We're setting up a second test fixture...                                                                                                                                                                   |
```
anselrognlie, line 121: we've seen quite a few fixture declarations now (with the addition of the fixture lesson) so I think we can be more matter-of-fact here
```suggestion
| `@pytest.fixture`                                   | Set up a second test fixture                                                                                                                                                                   |
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+| `def client(app):`                                  | This fixture is named `client`. It will request the existing `app` fixture to run, first.                                                                                                                   |
+| `return app.test_client()`                          | The responsibility of this fixture is to make a _test client_, which is an object able to _simulate_ a client making HTTP requests.                                                                         |
+
+### !callout-secondary
+
+## Practical Tip: Reading > Writing
+
+Understanding how to read this code is more valuable than learning how to write it. Realistically, on many projects, test environments are set up. However, it's valuable to read this code in order to:
```
anselrognlie, line 129: same edits as previous lesson with this callout
```suggestion
Understanding how to read this code is more valuable than learning how to write it. On many projects, we set up test environments by referring back to code from a previous project. But we should strive to be able to read this code in order to:
```

testing-our-apis/test-setup.md
```diff
@@ -0,0 +1,173 @@
...
+- Have a better understanding of how the app runs, which will lead to better debugging
+- Get familiar with the process of setting it up, in anticipation of future projects
+
+### !end-callout
+
+### The `client` Fixture
+
+The majority of our tests will use the `client` fixture that was defined in `tests/conftest.py`.
+
+While we're looking at `tests/conftest.py`, let's see a few ways we'll use `client`:
```
anselrognlie, line 140: we've been looking at conftest for a bit, so considering is appropriate, which means we can look at (intentional) the commands rather than just see (automatic) them
```suggestion
While we're considering `tests/conftest.py`, let's look at a few ways we'll use `client`:
```

anselrognlie (commented):
comments for test-setup

testing-our-apis/activity-solar-system-api-tests.md
```diff
@@ -0,0 +1,63 @@
...
+- Meet with your pair/group 👋
+- Decide which members will take the driver and navigator roles. This should be a switch compared to the last activity.
+
+The new driver may need to setup their database. Follow these steps:
+
+1. Pull down all new Git commits
+1. Activate the virtual environment
+1. Install the dependencies
+1. Create the database `solar_system_development`
+1. Create the `planet` table with the columns `id`, `name`, and `description`, and the one additional column
```
anselrognlie, line 26: this will be done by the upgrade command
```suggestion
```

testing-our-apis/activity-solar-system-api-tests.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Driver/Navigator Roles
+
+Driver and navigator roles should switch at least daily.
+
+### Build
+
+Fulfill the following requirements, with similar functionality to the Hello Books API:
+
+1. Create an `.env` file.
```
anselrognlie, line 38: "a" to match ".env" being read as "dotenv"
```suggestion
1. Create a `.env` file.
```

testing-our-apis/activity-solar-system-api-tests.md
```diff
@@ -0,0 +1,63 @@
...
+1. Populate it with two environment variables: `SQLALCHEMY_DATABASE_URI` and `SQLALCHEMY_TEST_DATABASE_URI`. Set their values to the appropriate value.
```
anselrognlie, line 39: repetitive use of value, but we can be more specific anyway
```suggestion
1. Populate it with two environment variables: `SQLALCHEMY_DATABASE_URI` and `SQLALCHEMY_TEST_DATABASE_URI`. Set their values to the appropriate connection strings.
```

testing-our-apis/activity-solar-system-api-tests.md
```diff
@@ -0,0 +1,63 @@
...
+1. Create a test database with the correct, matching name.
+1. Refactor the `create_app` method to:
+   1. Check for a configuration flag
+   1. Reads the correct database location from `.env`
```
anselrognlie, line 43: correct verb form
the os.environ is more accurately described as retrieving environment variables, and since we referred to the connection strings as environment variables in the previous step, using environment variable here is both more accurate and more consistent.
```suggestion
   1. Read the correct database location from the appropriate environment variables
```

anselrognlie (commented):
comments for activity-solar-system-api-tests

testing-our-apis/problem-set-testing-our-apis.md
```diff
@@ -0,0 +1,45 @@
...
+- Your completed repo
+- A list of any questions you weren't able to research independently
+
+## Practice
+
+Continue your independent Hello Books API.
+
+Fulfill the following requirements, with the same functionality presented in this topic:
+
+1. Create an `.env` file.
```
anselrognlie, line 20: same change as activity
```suggestion
1. Create a `.env` file.
```

testing-our-apis/problem-set-testing-our-apis.md
```diff
@@ -0,0 +1,45 @@
...
+1. Populate it with two environment variables: `SQLALCHEMY_DATABASE_URI` and `SQLALCHEMY_TEST_DATABASE_URI`. Set their values to the appropriate value.
```
anselrognlie, line 21: same change as activity
```suggestion
1. Populate it with two environment variables: `SQLALCHEMY_DATABASE_URI` and `SQLALCHEMY_TEST_DATABASE_URI`. Set their values to the appropriate connection strings.
```

testing-our-apis/problem-set-testing-our-apis.md
```diff
@@ -0,0 +1,45 @@
...
+1. Create a test database with the correct, matching name.
+1. Refactor the `create_app` method to:
+   1. Check for a configuration flag
+   1. Reads the correct database location from `.env`
```
anselrognlie, line 25: same change as activity
```suggestion
   1. Read the correct database location from the appropriate environment variables
```

anselrognlie (commented):
comments for problem set

config.yaml
```diff
@@ -1,29 +1,37 @@
 ---
 Standards:
   -
-    Title: Intro To Web Development
-    Description: "Web Applications and Layers, Intro to Clients, Servers, HTTP, Sequence Diagrams, Activity: Layers"
-    UID: 87ee542387da94693acb5efb099b7f58
+    Title: Testing Our Apis
+    Description: "Environments and Env Vars, dotenv and the Test DB, Test Setup, Using Tests, Problem Set: Testing Our APIs, Activity: Solar System API: Tests"
```
anselrognlie, line 5:
```suggestion
    Description: "Environments and Env Vars, dotenv and the Test DB, Fixtures, Test Setup, Using Tests, Problem Set: Testing Our APIs, Activity: Solar System API: Tests"
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+| `def test_get_all_books_with_no_records( ... ):`    | Continuing our best pytest practices, this test should start with the name `test_`, and it should describe the nature of this test |
+| `..._no_records(client):`                           | We pass in the `client` fixture here. pytest will automatically search and detect the fixture with the same name                   |
+| `client.get("/books")`                              | This sends an HTTP request to `/books`. It returns an HTTP response object, which we store in our local variable `response`        |
+| `response_body = response.get_json()`               | We can get the JSON response body with `response.get_json()`                                                                       |
+| `assert response.status_code == 200`                | Every `response` object will have a `status_code`. We can read that status code and check it against the expected status code.     |
+| `assert response_body == []`                        | We can check all of the parts of response body that we need to verify. We can check it's contents, size, values, etc!              |
```
anselrognlie, line 90:
```suggestion
| `assert response_body == []`                        | We can check all of the parts of the response body that we need to verify. We can check its contents, size, values, etc!              |
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+Add this test to `tests/test_routes.py`, and now let's run them!
```
anselrognlie, line 92: so far only one test
```suggestion
Add this test to `tests/test_routes.py`, and let's run it!
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+## Running Tests
+
+To run the tests, we use:
```
anselrognlie, line 96: emphasize that we run this the same way as other tests
```suggestion
To run our Flask tests, we can use the same methods as for any other test. From the command line we can run:
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+```bash
+(venv) $ pytest
+```
+
```
anselrognlie, line 101: give vs code as an option
```suggestion
Or we can run them through the VS Code test panel.

```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+We should see this already-implemented route pass!
+
+![Screenshot of pytest test result: 1 test passed](../assets/testing-apis/testing-apis_passing-get-books.png)
+
+Notice that we do _not_ need to run the Flask server to run the tests!
```
anselrognlie, line 106: this is ok, but I feel like since we're talking about something that just happened, we should use past tense
```suggestion
Notice that we did _not_ need to run the Flask server to run the tests!
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+### Verify Broken Tests Can Fail
+
+We haven't seen a failing test yet, because we are writing tests for already built functionality.
+
+Let's verify that we can trust our tests to fail, by making them fail temporarily.
+
+Let's go to our `app/routes.py` file and temporarily break our `/books` endpoint. One way we can break this route is to return a response with a status code `418` immediately.
```
anselrognlie, line 114: I feel like immediately on its own is sort of floating there, like does it mean we should make the change right now, or should that be the first line of the function (we want the latter meaning, but it feels more like the former)
```suggestion
Let's go to our `app/routes.py` file and temporarily break our `/books` endpoint. One way we can break this route is to return a response with a status code `418` before doing anything else in the function.
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+```python
+@books_bp.route("", methods=["GET", "POST"])
+def books():
```
anselrognlie, line 118: update according to function renaming
```suggestion
def handle_books():
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+    return Response("This is one broken response!", status=418)
```
anselrognlie, line 119: Let's go all in on I'm a teapot. It might get the reader to search for what's going on with this code, which is amusing. The subsequent text in the lesson isn't affected by this change.
```suggestion
    return Response("I'm a teapot!", status=418)
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+However, this test isn't complete yet...
+
+If we run this test right now, we actually get a 404!
+
+![Screenshot of pytest test result: test_get_one_book fails because of AssertionError, which checks if 404 is equal to 200](../assets/testing-apis/testing-apis_404-get-books-1.png)
+
+Why would we get a `404` response?
+
```
anselrognlie, line 166:
```suggestion

<br />

```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+ -->
+
+- First, let's confirm we understand what the test is doing:
+
+```python
+def test_get_one_book(client):
+    response = client.get("/books/1")
+```
+
+Our test made a `GET` request to `/books/1`. A `GET` to `/books/1` can only mean that our test made a request to our `book()` route.
```
anselrognlie, line 189: update for function renaming
```suggestion
Our test made a `GET` request to `/books/1`. A `GET` to `/books/1` matches the `/books/<book_id>` route with the `GET` method. This is mapped to our `handle_book()` function.
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+- Let's visit our `book()` route:
```
anselrognlie, line 191: update for function renaming
```suggestion
- Let's revisit our `handle_book()` function:
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+```python
+@books_bp.route("/<book_id>", methods=["GET", "PUT", "DELETE"])
+def book(book_id):
```
anselrognlie, line 195: update for function renaming
```suggestion
def handle_book(book_id):
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+    book = Book.query.get(book_id)
+    if book == None:
+        return Response("", status=404)
+    # ... rest of our route
+```
+
+Our `book()` route returns a `404` response if `book` is `None`.
```
anselrognlie, line 202: update for function renaming
```suggestion
Our `handle_book()` function returns a `404` response if `book` is `None`.
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+- Let's consider why `book` might have a value of `None`.
+  1.  What is the most line of code that most recently affected `book`?
```
anselrognlie, line 205:
```suggestion
  1.  What is the line of code that most recently affected `book`?
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+  1.  `book` is assigned a value in the line before, `book = Book.query.get(book_id)`.
+  1.  `Book.query.get(book_id)` must have returned `None`.
+
+Why would `Book.query.get(book_id)` return `None`? What are the reasons it behaved that way?
```
anselrognlie, line 209: consider posing this in a more investigative tone
```suggestion
Why would `Book.query.get(book_id)` return `None`? What could cause it to behave that way?
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+- Let's dive into `Book.query.get(book_id)`.
```
anselrognlie, line 211: we don't really get to see inside of this method, so phrase this a little differently
```suggestion
- What do we know about how `Book.query.get(book_id)` works?
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+  1.  The responsibility of this method is to return an instance of `Book` that has the primary key of `book_id`.
+  1.  This method returns `None` when no `Book` with that id was found!
+
+Our debugging questions can continue in this line of thought:
```
anselrognlie, line 215:
```suggestion
Our debugging questions can continue along this line of thought:
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+1. Why is there no book that was found?
+1. Where is the test looking for books?
+1. Is the test (and the `app` object made in `create_app()`) set to look at the correct database?
+1. Are our environment variables accessed and set correctly?
+1. Can we access our test database and check the contents of it in `psql`?
+1. Is what we see inside our test database consistent with the test error?
```
anselrognlie, line 222: this is going require some discussion and because the database contents are dropped after each test, checking the db state is tough... So generalize this. The next change will address how to draw the final conclusion.
```suggestion
1. Are there any books in the test database?
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+At this point, we should see that our test database has no book records inside of it.
```
anselrognlie, line 224: just looking in the db won't provide any useful info at this point, so walk through the thought experiment that will allow us to understand why the test is failing.
```suggestion
This final question is harder to confirm than might be expected! In the Test Setup lesson, we very briefly looked at the `app` fixture. But the full significance of the `db.create_all()` and `db.drop_all()` lines within the fixture may not have been apparent.

At the start of _each_ test, `db.create_all()` will rebuild the entire test database, creating all the tables needed for our models. Crucially, these tables begin _empty_!

Then at the end of _each_ test, `db.drop_all()` deletes _all_ the tables from the database. This prevents any test data created during one test from contaminating later tests. But this also means that if a test is not actively running, the test database will appear to be _completely_ empty! We won't even see empty tables. There will be _no_ tables!

Using our ability to debug tests with breakpoints though VS Code, it _is_ possible to pause during the middle of a test, which would give us a chance to connect to the test database and check on its state. This can be very helpful for more complicated tests!

But for the time being, we need only apply the knowledge that `db.create_all()` recreates the test database tables in an empty state at the start of each test.

This means that when we try to get the `Book` with a primary key of `1`, no `Book` is found. `Book.query.get(book_id)` returns `None`, resulting in the `404 Not Found` response status code and the failure of our test!
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+</details>
+
+To address our test failure, let's see one way to populate the test database.
```
anselrognlie, line 228: looking at is more active than seeing
```suggestion
To address our test failure, let's look at one way to populate the test database!
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+## Adding Test Data with Fixtures
+
+We'll define parts of our "Arrange" sections inside of pytest fixtures.
```
anselrognlie, line 232: mention other methods we could approach this, but then give a reason why fixtures is a better fit here
```suggestion
We could write code directly in our test to create data as part of our "Arrange" section. We could write helper functions if this data needs to be shared with other tests.

Or by using fixtures instead, we get data reusability and we explicitly list the data dependencies our test has!
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+Let's consider how we'll use fixtures to add test data:
```
anselrognlie, line 234: rephrase to acknowledge that we have discussed the generalities of fixtures already.
```suggestion
Let's review some of the characteristics of fixtures, and consider how we can use them to add test data:
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+| Question              | Answer                                                                                                                                         |
```
anselrognlie, line 236: wrapping format
```suggestion
| <div style="min-width:100px;">Question</div>              | Answer                                                                                                                                         |
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+| --------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
+| What are fixtures?    | Each pytest fixture we define will describe something we want to happen before the test, such as saving test data.                             |
```
anselrognlie, line 238: future sounds like we haven't seen how this works at all yet, but we have. so present sounds more general.
```suggestion
| What are fixtures?    | Each pytest fixture we define describes something we want to happen before a test. We could define a fixture that creates test data.                             |
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+| Where do fixtures go? | The fixtures can be defined in any file (such as `tests/test_routes.py`), but we will define most of our routes inside of `tests/conftest.py`. |
```
anselrognlie, line 239: routes → fixtures
give some minimal justification for why we're putting our fixtures there 
```suggestion
| Where do fixtures go? | Fixtures can be defined in any file (such as `tests/test_routes.py`), but we will define most of our fixtures inside of `tests/conftest.py`. This automatically allows them to be used in multiple test files if needed. |
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+| Who uses them?        | Each test will state which fixtures they want to use, and each test can use multiple fixtures!                                                 |
```
anselrognlie, line 240: now each of the previous rows is following a two-sentence pattern, so continue that here. I think this also helps emphasize the multiple fixture part.
```suggestion
| Who uses them?        | Each test states which fixtures they want to use. Each test can use multiple fixtures!                                                 |
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+    # Alternatively, we could do
+    # db.session.add(ocean_book)
+    # db.session.add(mountain_book)
+    db.session.commit()
+```
+
+| <div style="min-width:250px;"> Piece of Code </div> | Notes                                                                                                         |
+| --------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- |
+| `@pytest.fixture`                                   | Each fixture starts with this decorator                                                                       |
+| `def two_saved_books(...):`                         | We can name our fixtures whatever we want. `two_saved_books` is good enough, although we could think of more! |
```
anselrognlie, line 265: what makes it good enough?
```suggestion
| `def two_saved_books( ... ):`                         | We can name our fixtures whatever we want. `two_saved_books` is a reasonable name that will help remind us what this fixture does. |
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+| `def two_saved_books(app):`                         | This fixture needs to request to use the `app` fixture, defined above, so we have access to our Flask app     |
```
anselrognlie, line 266: follow previous explanations that list the params alone
the app fixture is in a previous lesson, not above in this lesson
we don't actually use the app param, but it's still important and we can say why (interestingly, because of the fixture param ordering, we can leave off app and it will still work, but it's still probably better to list the explicit dependency)
```suggestion
| `app`                         | This fixture needs to request the use of the `app` fixture, defined previously, so we know the test database has been initialized.     |
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+| `ocean_book = ...`                                  | We can make our first `Book` instance...                                                                      |
+| `mountain_book = ...`                               | ... and our second `Book` instance                                                                            |
+| `db.session.add_all([ ... , ... ])`                 | We can use the `add_all()` function to add a list of instances                                                |
+| `db.session.commit()`                               | This line commits and saves our book to the database                                                          |
```
anselrognlie, line 270:
```suggestion
| `db.session.commit()`                               | This line commits and saves our books to the database                                                          |
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+
+#### Requesting Fixtures in Tests
+
+To actually use this fixture in a test, we need to request this fixture by name.
```
anselrognlie, line 274: repeated "this fixture" sounds odd, and since we've had a whole lesson on fixtures, I don't think we need the repetition here.
```suggestion
To actually use this fixture in a test, we need to request it by name.
```

testing-our-apis/using-tests.md
```diff
@@ -0,0 +1,348 @@
...
+E         Differing items:
+E         {'title': 'Ocean Book'} != {'title': ''}
+E         {'id': 1} != {'id': ''}
+E         {'description': 'watr 4evr'} != {'description': ''}
+E         Use -v to get the full diff
+```
+
+![Screenshot of pytest test result: test_get_one_book failed because of AssertionError between two book dictionaries](../assets/testing-apis/testing-apis_failing-fixture-comparison.png)
+
+We should conclude that we need to update our test itself, finally! Let's fill in the expected test dictionary back in `tests/test_routes.py`. It should be consistent with whatever data we put in our fixture.
```
anselrognlie, line 302: provide an explanation for how to interpret the test results
"should" sounds a little judgmental in the case that I the reader wasn't able to arrive at that conclusion on my own.
```suggestion
The output shows that the data coming back from our API looks reasonable, but that the value we are comparing to in our test has empty strings for each dictionary value.

We can conclude that we need to update our test itself, finally! Let's fill in the expected test dictionary values back in `tests/test_routes.py`. It should be consistent with whatever data we put in our fixture.
```

anselrognlie (commented):
comments for using-tests

anselrognlie (changes requested):
Finished review with comments.

Also, you'll want to review my uploaded fixture lesson.

testing-our-apis/dot-env-and-the-test-db.md
```diff
@@ -0,0 +1,193 @@
...
+
+`.env` files are formatted in the following way:
+
+- Each line is a different environment variable
+- Environment variables are named, separated by a `=`, and then given a value (with no spaces or quotes)
+  - It's conventional to name environment variables with CAPITAL_CASING, just like constant variables
+
+Beyond those guidelines, we are free to choose the best variable names and values for our project.
+
+For our Hello Books API, we want to designate a path to a development database and a testing database with the variables named `SQLALCHEMY_DATABASE_URI` and `SQLALCHEMY_TEST_DATABASE_URI`.
```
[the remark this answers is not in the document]
tildeee: Gonna bring this into an issues for a future refactor! I like the suggestion, it just happens to be a lot to refactor in one go. Thanks!

testing-our-apis/environments-and-env-vars.md
tildeee: YAYYYY staging!!! I wasn't brave enough to pull in staging but this is great!! Thanks!

testing-our-apis/dot-env-and-the-test-db.md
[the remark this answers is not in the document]
anselrognlie: Yeah, I definitely wouldn't change this right now.

testing-our-apis/environments-and-env-vars.md
anselrognlie: It _could_ get confused with the test environment, but I think the practice studio you provided in the metaphor will do a great job of helping the reader keep these things separate in their mind, even though we won't really be using staging environments.

testing-our-apis/fixtures.md
```diff
@@ -0,0 +1,324 @@
...
+| `@pytest.fixture`                                   | Applies the fixture decorator to the `empty_list` function. `pytest` will be able to use `empty_list` as a fixture in subsequent tests.                                                  |
+| `def empty_list(): ...`                             | Declares the `empty_list` function, which returns an empty list.                                                                                                                         |
+| `def test_len_of_empty_list( ... ): ...`            | Declares a test function that `pytest` will be able to detect. It begins with `test_`.                                                                                                   |
+| `empty_test`                                        | This parameter matches the name of the `empty_list` fixture. When `pytest` runs this test it will first run the `empty_list` fixture and use the result as the value for this parameter. |
+
+We can run this test file with `pytest test_fixtures.py`. If we are not in our Hello Books API environment, we may need to create a virtual environment and install `pytest` so that the `pytest` command is available.
```
tildeee, line 62: Alternatively,

```suggestion
We can run this test file with `python3 -m pytest test_fixtures.py`.
```

testing-our-apis/fixtures.md
```diff
@@ -0,0 +1,324 @@
...
+
+##### !explanation
+
+Here's an explanation for each option:
+
+- This option doesn't follow `pytest` naming conventions, so it will not be recognized as a test.
+- This option omits the `@pytest.fixture` decorator.
+- This option calls the fixture function directly. `pytest` disallows this and will raise an error.
+- This option is correct. It uses the decorator, follows `pytest` naming conventions, and declares the fixture as a dependency in its parameter list.
+- This option properly declares the fixture, but doesn't list it in the test parameter list.
```
tildeee, line 299: Suggest for this to be a numbered list for slightly greater ease of "following"/mapping to the above options

```suggestion
1. This option doesn't follow `pytest` naming conventions, so it will not be recognized as a test.
1. This option omits the `@pytest.fixture` decorator.
1. This option calls the fixture function directly. `pytest` disallows this and will raise an error.
1. This option is correct. It uses the decorator, follows `pytest` naming conventions, and declares the fixture as a dependency in its parameter list.
1. This option properly declares the fixture, but doesn't list it in the test parameter list.
```

testing-our-apis/fixtures.md
```diff
@@ -0,0 +1,324 @@
...
+| Fixture              | Shared code used to perform setup and cleanup for tests.                                     | Setup code, Cleanup code | "We're doing a lot of the same setup and cleanup in these tests. Let's refactor that into fixtures that can be shared among all of them!" |
+| Dependency           | In this context, a fixture listed as one of the parameters in a test function.               | -                        | "My test was failing because I forgot to list one of its dependencies in the parameter list."                                             |
+| Dependency Injection | A way for code to explicitly declare and receive the resources it needs to run successfully. | -                        | "My tests receive the resources they need because `pytest` automatically injects the dependencies."                                       |
+
+## Creating Basic Fixtures
+
+We create a fixture with the `@pytest.fixture` decorator.
+
+Let's look at a small test file. Let's run `touch test_fixtures.py` to create a new file, then paste in the following code.
```
tildeee, line 40: For the sake of making a decision for them, I may recommend being really concrete here:

```suggestion
Let's create a small test file to practice using fixtures and tests. We can create this small test file in any folder of notes. Inside our notes folder, run `touch test_fixtures.py` to create a new file, then paste in the following code.
```

config.yaml
```diff
@@ -1,29 +1,37 @@
 ---
 Standards:
   -
-    Title: Intro To Web Development
-    Description: "Web Applications and Layers, Intro to Clients, Servers, HTTP, Sequence Diagrams, Activity: Layers"
-    UID: 87ee542387da94693acb5efb099b7f58
+    Title: Testing Our Apis
```
tildeee, line 4:
```suggestion
    Title: Testing Our APIs
```

testing-our-apis/fixtures.md
tildeee:
```suggestion
Let's create at a small test file that we'll use to experiment for this lesson. We'll only use this test for this lesson, so we'll delete this file at the end of the lesson. Inside of our Hello Books API project, let's run `touch test_fixtures.py` to create a new file, then paste in the following code.
```

testing-our-apis/fixtures.md
tildeee:
```suggestion
Because we're in the Hello Books API project, let's activate our virtual environment. Then, let's run this test file with `pytest test_fixtures.py`.
```

anselrognlie: We will specifically direct the reader to create this file in Hello Books API.

testing-our-apis/fixtures.md
anselrognlie: Thanks! Made a few edits to your comment. (tried to add my own comment to preserve your original input, but then GH wouldn't let me include my comment in the batch?)

testing-our-apis/fixtures.md
anselrognlie: Thanks. Made some edits here too.

anselrognlie (commented):
just trying to get GH to stop thinking i'm in a review...

testing-our-apis/fixtures.md
```diff
@@ -0,0 +1,324 @@
...
+Generators and coroutines are powerful features in Python, but are beyond the scope of this curriculum. We will only use `yield` in the context of fixtures. But generators and coroutines can be an interesting topic for further research. Follow your curiosity!
+
+### !end-callout
+
+## Summary
+
+Fixtures let us move our setup and cleanup code into shared functions. These functions are clearly marked with the `@pytest.fixture` decorator so we can be confident that those functions are used for test setup and cleanup. We can declare that a test depends on a particular fixture by listing its name as a parameter to the test.
+
+The examples we looked at in this lesson aren't very "useful", but in real cases where we use fixtures, we can perform more complicated setup. We could initialize the resources needed to test our API endpoints, or set up or tear down a test database, as we'll see in the next lesson!
+
```
anselrognlie, line 218:
```suggestion

### !callout-info

## Let's Cleanup Our Test File!

We can delete the `test_fixtures.py` file from our project now! Or we can set it aside in another folder if we'd like to refer back to it later. But let's be sure to remove it from this project since it was only being used as an example for this lesson!

### !end-callout

```

PR #15, opened 2021-04-07
more-selecting-in-sql: tiny cfu questions for syntax ordering

tildeee:


anselrognlie (approved):
Looks good! Especially good job splitting the chunks so there was only one correct ordering.

PR #16, opened 2021-04-08
Ar/problem set more selecting

anselrognlie:
adds sql questions for more selecting in sql
- In order to test this, I had to deploy to a sandbox. It would not run properly in learn preview.
- I did confirm all the questions "work".
- Currently deployed to https://learn-2.galvanize.com/cohorts/2621/blocks/1310/content_files/more-selecting-in-sql/problem-set-more-selecting-in-sql.md
- in the summary, i listed the problem set before the optionals, but in the lesson, I still put the problem set at the end. let me know if this should be altered.


config.yaml
```diff
@@ -116,6 +116,10 @@ Standards:
         Type: Lesson
         UID: af18e4ef-18aa-432c-9f34-0375f4dbb82a
         Path: /more-selecting-in-sql/optional-grouping.md
+      -
+        Type: Lesson
+        UID: f9f13580-bd04-442a-8944-34864f3de6a2
+        Path: /more-selecting-in-sql/problem-set-more-selecting-in-sql.md
```
tildeee, line 122: Personally, I think this PS should be listed also before optionals (not just in the description)

more-selecting-in-sql/problem-set-more-selecting-in-sql.md
```diff
@@ -0,0 +1,322 @@
...
+```sql
+SELECT name
+FROM flowers
+WHERE soil_type = 'acidic'
+AND light_level = 'partial shade';
+```
+
+##### !end-tests
+
+### !end-challenge
```
tildeee, line 58: In my ideal, there'd be an explanation with all of these with an example solution (probably just a literal copy/paste of the test is fine)

```suggestion
##### !end-tests
##### !explanation

This is an example solution:

##### !end-explanation
### !end-challenge
```

tildeee (approved):
Awesome!!!!

I've left two comments that I'll redundantly say here:

1. I think you made the right call in the config/topic description; I say we move up the file too
2. I think adding a sample solution to each in an explanation would be cool

PR #17, opened 2021-04-09
building-an-api: refactors flask scaffold to remove app.py

tildeee:
Keeping it as a PR because I should check screenshots, check that the update and delete, 404 and more flask queries, env vars, and testing topics don't reference this file


jmaddox19 (approved):
Looks good to merge into the other PR to me! lol

PR #26, opened 2021-04-19
data-in-tables: Add questions to the review section of the activity

jmaddox19:


data-in-tables/instructor-activity-data-in-tables.md
```diff
@@ -3,3 +3,15 @@
 The goal is to get students to establish motives for using data in tables by thinking of realistic features they'd build
 
 This should help train a brain to go "What do I want to build? If I want to build that, what kind of data and table do I need?"
+
+## Notes for Review
+
+### "Why wouldn’t Craigslist just store this data in a list of dictionaries?"
+
+A good conclusion to offer to the class would be something like "We will come to realize there are several reasons for this but the main one we'll focus on right now is the need for persistence, for the data to live on when our program is finished running."
```
anselrognlie, line 11: At the time of doing this activity, students haven't technically been presented with the idea of a database as a means of persistence. This activity only asks them to think about arranging data. I think bringing this up in discussion is a good way to start turning their thinking towards Intro to SQL, but it feels like a big ask for this to be something that students will come up with on their own.

data-in-tables/activity-data-in-tables.md
```diff
@@ -107,7 +107,11 @@ Come up with...
 
 ## How to Review
 
-1. In Part 2, did the columns for a post in "services" differ from the columns for a post in "for sale"? In what ways?
+1. Why wouldn’t Craigslist just store this data in a list of dictionaries?
```
tildeee, line 110:
```suggestion
```

Agreeing with Ansel's comment, this question is a pretty high-level evaluation question. It feels like it's trying to get students to justify why craigslist would store data in tables (or possibly in a database which, as Ansel mentioned, hasn't been defined at this point). Similarly, I wouldn't expect students to be on the ["evaluation" level](https://en.wikipedia.org/wiki/Bloom%27s_taxonomy#Evaluation) of determining why craigslist _wouldn't_ store it in a list of dictionaries, making those "why" conclusions requires understanding and comparison.

data-in-tables/activity-data-in-tables.md
```diff
@@ -107,7 +107,11 @@ Come up with...
 
 ## How to Review
 
-1. In Part 2, did the columns for a post in "services" differ from the columns for a post in "for sale"? In what ways?
+1. Why wouldn’t Craigslist just store this data in a list of dictionaries? 
+1. Why would it be more work to add a new column to one of these tables than to add a row?
```
tildeee, line 111: I'm not sure what kind of answer you're looking for, besides repeating the callout that's in the "Data in Tables" activity, which is moreso a fact to recall.

I wouldn't expect students to be able to answer _why_, especially since this Tables lesson lives in the abstract and we haven't talked about database management systems.

```suggestion
1. Is it more work to add a new column to one of these tables, or to add a row to one of these tables?
```

Here's a suggestion for a way to phrase the question that's closer to the Knowledge/Understanding level.

By the way, what part of the activity would directly inform the answer to this question? I'm trying to figure out which part the student would look back on in order to find the answer for this question.

data-in-tables/activity-data-in-tables.md
```diff
@@ -107,7 +107,11 @@ Come up with...
...
+1. In Part 2, did the columns for a post in "services" differ from the columns for a post in "for sale"? In what ways? Can a couple groups share what they got for part 2.3?
```
tildeee, line 112:
```suggestion
1. In Part 2, did the columns for a post in "services" differ from the columns for a post in "for sale"? In what ways?
```

Since I'd like to avoid the curriculum being a script, would you mind if we left out this sharing prompt?

data-in-tables/instructor-activity-data-in-tables.md
```diff
@@ -3,3 +3,15 @@
...
+
+It may also be worth clarifying that a database table is different from a list of dictionaries in that each dictionary might look totally different. In a database table, there's an enforcement that each item looks the same, so it would be more like a list of instances of the same class (eg. instances of the Listing class).
```
tildeee, line 13: Echoing what Ansel said, technically the word "database" hasn't been introduced or defined for them yet; we haven't even stated that the tables that they're emulating are database tables.

If a student was completely following the curriculum and didn't know what a database was, I wouldn't expect them to be able to make this comparison or judgment.

We've also only talked about data with apps in the abstract sense in this lesson, without naming programming languages. Bringing in Python dictionaries/lists wouldn't be bad, but it'd be coming out of nowhere, since this lesson's content is entirely not about thinking about code

data-in-tables/instructor-activity-data-in-tables.md
jmaddox19: Hmm...fwiw, they did come up with it on their own in Scissors. And also I wouldn't say this is written in a way to imply that there's a problem if they don't know the answer and the instructor just needs to offer the answer.
But if you have a suggestion for wording it to make it more clear this is a stretch question, I'm happy to take that suggestion

data-in-tables/activity-data-in-tables.md
jmaddox19: I'm curious to understand what's wrong with inviting them into this more exploratory thinking in a safe space like classroom discussion where they don't need to have the correct answer and the instructor will offer the correct answer after they've thought about it.

data-in-tables/activity-data-in-tables.md
jmaddox19: I wouldn't say there's a part of the activity that would inform this. It just seemed like the right spot for it since its the first classroom time with databases, and students did have the answer in scissors today. But I'm open to the input that this belongs in tomorrow's or Wed's activity instead

data-in-tables/instructor-activity-data-in-tables.md
anselrognlie: In Rock, no one ventured in the direction of persistence. Persistence hasn't been something we've really dealt with yet aside from a lightning talk, so it feels like we're opening a whole can of worms on this.

It's an interesting talking point, but posing a question that's ostensibly about data structures but is actually about data persistence seems a little bait-and-switch. To me this feels like it would need more discussion than it could get without the background that subsequent material will provide.

For a group without sufficient background, providing the answer and moving on feels unsatisfying and likely to evoke a "what the heck was that about" response.

PR #29, opened 2021-04-20
aligns question text to lesson changes

anselrognlie:
Some of the lesson text was changed from predict to plan. This updates the question checkboxes to match the verbiage.


tildeee (approved):
thank you!!!!

PR #34, opened 2021-04-22
updates code style of existing hello-books-api curriculum

tildeee:


building-an-api/flask-setup.md
```diff
@@ -270,40 +265,6 @@ We should try to become familiar with the kinds of settings we might want to con
...
-
-    db.init_app(app)
-    migrate.init_app(app, db)
-
-    # import models (data objects) here
-
-    # import routes (blueprints) here
-
-    return app
-```
```
anselrognlie, line 297: So with this removal, are we basically saying that the student _must_  clone from GH to see what this looks like? Was this the main place a student following along would have a place to get the contents of the file?

I feel like with this content, there's a chance that a student could have decided to do things on their own, but without this, they'll be dependent on the repo.

anselrognlie (approved):
Looks good. I just had a question (in the comments) about the implication of no longer showing __init__.py.

building-an-api/flask-setup.md
tildeee: Yes! Good convo. I agree; this removal would leave a student dependent on a given init file in a repo.

I still feel okay about that since:

- In the context of the lesson, I personally felt that it was emphasizing this file a lot
- The lessons are built as _kind of_ a walkthrough to build it themselves, with the motivations for each step. I think a student would get more out of following the walkthrough config steps and reading through the explanations, rather than looking here. ... It is a lot more text to shuffle through though lol

[The Hello Books API scaffold has a pretty bare init file](https://github.com/AdaGold/hello-books-api/blob/00-scaffold/app/__init__.py)

PR #49, opened 2021-09-20
changes the activity to design solar system api for future assignments

spitsfire:


beccaelenzil (commented):
I remember our conversation about moons and planets, but now I'm a bit confused about how this exercise follows the lesson. It seems here that we're suggesting students make nested routes, but this is not covered in the lesson, and establishing relationships comes a bit later in the Unit 2. Also, I remember talking about updating a planet by planting a flag, but again this seems like a custom route. Maybe we should just have them create CRUD routes for planets and moons at this stage, and think about their connection later?

PR #50, opened 2021-09-20
changes activity to add solar system api to part 4

spitsfire:


beccaelenzil (commented):
This looks good, but like the API Design exercise, I have question about how moons and planets relate, and if relationships are too early to bring in in the first week week of the Unit

PR #52, opened 2021-09-28
Be/building an api revision

beccaelenzil:
Changes were needed across multiple files to account for the fact that students will be refactoring the Read routes that they wrote with hardcoded data. The biggest changes are in read.md and create.md

I decided to leave some of the planning endpoints material as a review, but am on the fence about this decision. Feedback welcome.


building-an-api/models-setup.md
```diff
@@ -47,23 +47,50 @@ In a previous lesson we used hardcoded data stored in the variable `BOOKS` to le
 
 In this lesson, we will learn how to connect our Flask app to a database. This will allow us to not only write routes to *read* our data, but also *create*, *update*, and *delete* it!
 
-Before we go any further, let's remove our hardcoded `BOOKS` data from the `app`. 
+Before we go any further, let's comment our hardcoded `BOOKS` data and our two routes from the `app`. 
 
-This is the code to remove:
+This is the code to comment:
```
CheezItMan, line 52: Do we want them to comment out the code or remove it knowing they have an older commit to look back on if they want to.

CheezItMan (approved):
Overall I think this looks good.

building-an-api/models-setup.md
beccaelenzil: First I thought remove, then comment, but I think you have a good point. So maybe stick with remove?

building-an-api/models-setup.md
```diff
@@ -47,23 +47,50 @@ In a previous lesson we used hardcoded data stored in the variable `BOOKS` to le
 
 In this lesson, we will learn how to connect our Flask app to a database. This will allow us to not only write routes to *read* our data, but also *create*, *update*, and *delete* it!
 
-Before we go any further, let's remove our hardcoded `BOOKS` data from the `app`. 
+Before we go any further, let's comment our hardcoded `BOOKS` data and our two routes from the `app`.
```
beccaelenzil, line 50:
```suggestion
Before we go any further, let's comment or remove our hardcoded `BOOKS` data and our two routes from the `app`. 
```

PR #77, opened 2021-11-02
Fix deployment push instructions.

alope107:
- Our instructions previously assumed there was a branch named
`master` that the students had been using for hello-books-api.
With our current repositroy, this is incorrect. This commit
adds instructions on how to determine which branch the student is
on and how to use that branch for the push to heroku.

### DIFF
--- a/deployment/deploying-to-heroku.md
+++ b/deployment/deploying-to-heroku.md
@@ -197,10 +197,16 @@ Visit the Heroku dashboard and see your new app listed! We'll visit this dashboa
 
 We should send our project codebase to our `heroku` remote.
 
-This command will push our project's Git history to a remote named `heroku`. It will push the default `master` branch on our computer to Heroku's default `main` branch on our new Heroku virtual computer.
+First, we need to determine which branch our local repository is on. Run `git status` to get this information. The first line displayed by this command should be
+
+```
+On branch your-branch-name
+```
+
+We next will push our project's Git history to a remote named `heroku`. We will push the branch on our computer to Heroku's default `main` branch on our new Heroku virtual computer. In the command below, replace `your-branch-name` with the name of the branch you identified using `git status`.
 
 ```bash
-$ git push heroku master:main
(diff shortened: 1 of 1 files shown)


deployment/deploying-to-heroku.md
```diff
@@ -197,10 +197,16 @@ Visit the Heroku dashboard and see your new app listed! We'll visit this dashboa
...
 We should send our project codebase to our `heroku` remote.
 
-This command will push our project's Git history to a remote named `heroku`. It will push the default `master` branch on our computer to Heroku's default `main` branch on our new Heroku virtual computer.
+First, we need to determine which branch our local repository is on. Run `git status` to get this information. The first line displayed by this command should be
+
+```
+On branch your-branch-name
+```
+
+We next will push our project's Git history to a remote named `heroku`. We will push the branch on our computer to Heroku's default `main` branch on our new Heroku virtual computer. In the command below, replace `your-branch-name` with the name of the branch you identified using `git status`.
```
anselrognlie, line 206: "virtual computer" sounds a little out of left field to me. I could see virtual machine, or even dynamo (to bring in some of heroku's terminology). But as we haven't really introduced any of those terms very well, and so as to avoid drawing unnecessary attention to itself, how about using environment?

```suggestion
We next will push our project's Git history to a remote named `heroku`. We will push the branch on our computer to Heroku's default `main` branch in our new Heroku environment. In the command below, replace `your-branch-name` with the name of the branch you identified using `git status`.
```

anselrognlie (approved):
This seems ok for now (minor wording suggestion), though it would be nice to more completely explain what the 'local:remote' part is doing. When we ran into this in C15, lost of students ended up `git push heroku main:main` which can really be written as `git push heroku`. But this fix will work for both task-list (which has a master branch) and hello-books/solar-system (which don't).

alope107:
Agreed that more explanation would be ideal here. It's a good idea to add that for C17.

PR #80, opened 2022-03-14
Be/expire all

beccaelenzil:
In C16 we discovered that some of our tests didn't actually test whether changes were persisted in the database. 

Adding db.session.remove() to the conftest is one way to address this. A bit more info here: https://stackoverflow.com/questions/39480914/why-db-session-remove-must-be-called

The exact details of how this works feels beyond the scope of our curriculum. 
I'd love feedback on the way description provided for `db.session.remove()` in the table.

Here is the proposed change to tasklist: https://github.com/AdaGold/task-list-api/pull/21
And to hello-books-api: https://github.com/AdaGold/hello-books-api/pull/3

Once we're confident on this change, we can propagate the change to the hello-books branches and make any changes to the video lessons that are needed.


CheezItMan (approved):
Looks good!

testing-our-apis/test-setup.md
```diff
@@ -117,9 +122,12 @@ def client(app):
...
+|`from flask.signals import request_finished`|We use the `@request_finished` decorator to create a new database session after a request as described below. | 
 | `@pytest.fixture`                                   | We'll create and use a pytest fixture named `app`, which will be used in our `client` fixture (defined later)                                                                                              |
 | `app = create_app( ... )`                           | When we run our tests, this line will run and create an `app` object. It's using the same `create_app` function defined in our `app/__init__.py` file!                                                      |
 | `{"TESTING": True}`                                 | Here, we're passing in a dictionary to represent a "test config" object. The current implementation of `create_app()` in `app/__init__.py` uses this argument only to check if it's truthy (which this is!). |
+|`@request_finished.connect_via(app)`| This decorator indicates that the function defined below `expire_all` will be invoked after any request is completed |
```
anselrognlie, line 129: Does `expire_all` appear elsewhere in the example?  From the context here, I don't know what it's supposed to mean.

anselrognlie (commented):
I'm confused by the description of `@request_finished.connect_via(app)`

testing-our-apis/test-setup.md
anselrognlie: Maybe this was the intent?

```suggestion
|`@request_finished.connect_via(app)`| This decorator indicates that the function defined below, `expire_session`, will be invoked after any request is completed |
```

testing-our-apis/test-setup.md
```diff
@@ -117,9 +122,12 @@ def client(app):
...
+|`db.session.remove()`| After a request is made in our test, this line creates a new database session so that we can test that changes were persisted in the database. This is particularly relevant for testing the update method. Without this line, when we test that an update was made to a record following a `put` request, the test will only look at the local copy of the updated record. By adding `db.session.remove()`, we make sure the test checks that the update was persisted in the database.
```
anselrognlie, line 130: Suggest in-memory rather than local. For most of the projects, the db students use will also be on their local machine, so local is a little ambiguous.

```suggestion
|`db.session.remove()`| After a request is made in our test, this line creates a new database session so that we can test that changes were persisted in the database. This is particularly relevant for testing the update method. Without this line, when we test that an update was made to a record following a `put` request, the test will only look at the in-memory copy of the updated record. By adding `db.session.remove()`, we make sure the test checks that the update was persisted in the database.
```

anselrognlie (commented):
A couple small suggestions. Otherwise, the level of detail feels about right to me.

testing-our-apis/test-setup.md
beccaelenzil: yes! I'm not sure where I got the name `expire_all` from

PR #101, opened 2022-03-31
Deployment and Solar System update

beccaelenzil:
- Updated Deploying to Heroku format
- Added to the callout in Intro to Hello Books
- Changes Solar System Activities to type: Lesson so instructors can decide whether to make it visible or not


CheezItMan (approved):
So you decided to default to making the Solar System instructions in Building an Api visible?

PR #108, opened 2022-04-02
Updates to many-to-many lessons

beccaelenzil:
I accidentally made commits to `main`, but given that this material is not live yet, I thought it was okay to make the  PR in reverse order (`main` has the changes to the Many-to-Many lessons and `be/api-09` was branched off before those changes were made)


api-9-many-to-many/join-table-many-to-many.md
```diff
@@ -0,0 +1,155 @@
+# BookGenre Join Table
+
+## Goals
+
+Our goal for this lesson is to connect the `Book` and `Genre` models to create a many-to-many relationship.
```
CheezItMan, line 5: I feel like this should be a learning goal rather than a specific feature of the hello books app.

CheezItMan (approved):

PR #109, opened 2022-04-07
change exiting to existing and add return to code table

beccaelenzil:
just noticed that prettier reformatted tables!


anselrognlie (approved):
Looks good to me!

PR #124, opened 2022-08-16
Fix use of HTML-like tag within <code> tags.

marciaga:
When reviewing course content, I noticed a double slash where a variable should have gone, i.e.
```
<code>POST /genres/<genre_id>/books</code>
```
The formatter interprets `<genre_id>` as a tag and doesn't render the text, so I changed it to use HTML entities so the rendering looks correct. 

I also fixed a path typo, which used \path\to\thing instead of /path/to/thing

Here's a link to the preview: https://learn-2.galvanize.com/cohorts/3441/blocks/452/content_files/api-9-many-to-many/routes-many-to-many.md

### DIFF
--- a/api-9-many-to-many/routes-many-to-many.md
+++ b/api-9-many-to-many/routes-many-to-many.md
@@ -81,7 +81,7 @@ Our route function we will need to
 <br/>
 
 <details>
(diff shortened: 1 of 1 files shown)


anselrognlie (commented):
LGTM

anselrognlie (approved):
LGTM

PR #129, opened 2022-10-20
Update code examples in Building an API pts. 8-10

kelsey-steven-ada:
Changes:
- Incorporate refactoring changes from pt. 7 of Building an API series into the code examples in parts 8-10. 
- In pt. 10 routes-many-to-many.md, updates language around the `to_dict` function to reflect that we want to update the existing function from pt. 7
- Also fixes a couple issues with angle brackets inside `<code>` tags causing the text to disappear


kyra-at-ada (approved):
I'm not sure what it is about this repo, but I still get the bug where I can't preview the lesson.

Other than that, it looks good! Thanks for making these updates!

PR #143, opened 2024-04-10
Building an API Pt. 3 - Lesson 1 Updates

kelsey-steven-ada:
[Asana ticket](https://app.asana.com/0/1205655776549324/1207004723163599/f)

Changes:
- Comment out videos & add marker to find the commented out section in the future. 
- Update the project structure diagram to reflect new folders. 
- Update SQLAlchemy docs link

### DIFF
--- a/api-3-database-models-read/models-migrations.md
+++ b/api-3-database-models-read/models-migrations.md
@@ -1,10 +1,11 @@
 # Introduction to Models and Migrations
 
-<iframe src="https://adaacademy.hosted.panopto.com/Panopto/Pages/Embed.aspx?pid=d1fcd890-3fb6-404d-b28f-ad110031a865&autoplay=false&offerviewer=true&showtitle=true&showbrand=false&start=0&interactivity=all" height="405" width="720" style="border: 1px solid #464646;" allowfullscreen allow="autoplay"></iframe>
+<!-- FLASK UPDATE -->
+<!-- <iframe src="https://adaacademy.hosted.panopto.com/Panopto/Pages/Embed.aspx?pid=d1fcd890-3fb6-404d-b28f-ad110031a865&autoplay=false&offerviewer=true&showtitle=true&showbrand=false&start=0&interactivity=all" height="405" width="720" style="border: 1px solid #464646;" allowfullscreen allow="autoplay"></iframe> -->
 
-## Goal
+## Goals
 
-Our goal for this lesson is:
+Our goals for this lesson are to:
 
 1. Introduce the vocabulary "model," as it's commonly used in web development, and will be used in Flask
 1. Briefly introduce the steps to setting up a model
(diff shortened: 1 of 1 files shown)


api-3-database-models-read/models-migrations.md
anselrognlie, line 1: Since we started out by placing the non-db-backed Book class in the models directory, and presumably had at least some text earlier on describing why that's where we put it (in anticipation of this material), would it be worth reworking the `Models in Flask` or `Models Are a Link Between a Database and Code ` sections (maybe it would need to be a little non-localized?) to bridge from our plain conceptual model to our soon-to-be-db-backed model?

As a reader, I feel like the text doesn't take into account that we already have a class sitting there. Like later it's talking about some hypothetical model, when we already have an actual model.

Maybe tone down some of the language that draws a hard connection between models and the database? Either that, or make it clear that while a model can generally refer to any of the core abstractions in our business logic, that moving forward in our work with Flask, we will primarily use model to refer to those entities that are backed by the database?

api-3-database-models-read/models-migrations.md
```diff
@@ -95,7 +99,7 @@ Generating and applying these instructions in separate steps allows for teams to
 
 We don't need to write the code for storing, retrieving, or updating our models. SQLAlchemy lets us define our models as normal Python classes. Instead of writing full lines of SQL ourselves, we will write Python classes following patterns defined by SQLAlchemy, and SQLAlchemy will do most of the SQL heavy lifting for us!
 
-SQLAlchemy has [already defined methods](https://flask-sqlalchemy.palletsprojects.com/en/2.x/queries/) for:
+SQLAlchemy has [already defined methods](https://flask-sqlalchemy.palletsprojects.com/en/3.1.x/queries/) for:
```
anselrognlie, line 102: Do we need to link this? We're linking to the Flask-SqlAlchemy docs rather than the main SqlAlchemy, and while the insert, update, delete, select stuff is fine, that page also shows the `action_or_404` methods, which I would actively discourage students from using (since they produe an html error). The raw SqlAlchemy docs are probably a bit overwhelming to drop in here, but can't the reader just take our word for it that such methods exist, since we'll be covering the core of them in the coming lessons.

anselrognlie (commented):
Looks good and is fine as is, but I added some thoughts about the voicing given the project structure changes, and had a question about the documentation link.

api-3-database-models-read/models-migrations.md
kelsey-steven-ada: That makes a lot of sense, it's a weird tone to act as if we are freshly going to add to the models folder when we've already been working with a class there. 

I like adding in the reinforcement in the latter suggestion that models can be used more broadly, but we're going to be referring specifically to data-backed models as we move through the series.

anselrognlie (approved):
Love this. Those updates really help this read more naturally to me. And you were surgeon-like in your precision!

PR #144, opened 2024-04-10
API Lesson 1 Pt 3: Update lesson to reflect latest version of Hello-Books-API 

yangashley:
Asana Task: https://app.asana.com/0/1205655776549324/1207004723163589/f

Updates directory structure to include separate routes directory, fixes typos, updates sample requirements.txt with update package versions


api-1-setup-read/flask-setup.md
```diff
@@ -67,39 +68,26 @@ These files will list the exact packages and their version numbers to download a
...
+Mako==1.3.2
+MarkupSafe==2.1.5
+packaging==23.2
+pluggy==1.4.0
+psycopg2-binary==2.9.9
+pytest==8.0.0
+python-dotenv==1.0.1
+SQLAlchemy==2.0.25
+typing_extensions==4.9.0
+Werkzeug==3.0.1
```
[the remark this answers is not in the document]
kelsey-steven-ada: That looks good, better to show the new versions over older/possibly deprecated versions.

api-1-setup-read/flask-setup.md
```diff
@@ -201,14 +189,15 @@ This curriculum will provide a suggested project structure and location:
...
+│   └── routes
+│       └── __init__.py
 ├── README.md
 └── requirements.txt
 ```
 
 ### Routes
 
-Inside the `app` folder, there will be a file named `routes.py`. The responsibility of this file is to define the endpoints.
+The `app/routes` directory will be responsible for holding our routes. We should organize our routes so that all the routes corresponding to a particular model will be written in a single file. For example, all the endpoints related to the Book model would be written in a file called `book_routes.py`.
```
kelsey-steven-ada, line 200: Not strictly necessary, but I was thinking the class name should be formatted as well.

```suggestion
The `app/routes` directory will be responsible for holding our routes. We should organize our routes so that all the routes corresponding to a particular model will be written in a single file. For example, all the endpoints related to the `Book` model would be written in a file called `book_routes.py`.
```

api-1-setup-read/flask-setup.md
```diff
@@ -23,7 +24,7 @@ We will:
 
 ## Manage Dependencies
 
-Whenever we work on a Python project, we need to consider how we manage our dependencies, the third party packages used by our project.
+Whenever we work on a Python project, we need to consider how we manage our dependencies which are the third party packages used by our project.
```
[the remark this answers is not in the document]
anselrognlie: But by the  time students see this section, they will be using python 3.12, which will prevent them from pip installing system-wide at all, and we'll have started students using venvs as "projects" in precourse (and they'll have seen it on all the unit 1 projects), so I think this segue text could do with a re-flow. The main things for it to touch on should be:
1. Flask projects are Python projects, just like the other projects we've done.
2. We use venvs to manage our Python projects.
3. We will use a venv to manage our Flask project.
4. Here's a reminder on how to set up a venv for a project when the requirements are supplied to us.

api-1-setup-read/flask-setup.md
```diff
@@ -220,7 +209,7 @@ Inside each `app` folder, there will be a file named `__init__.py`. This is the
 
 The start-up logic is responsible for locating and applying any app configuration, and getting the server ready to receive requests.
 
-To get started, we will include the following code in `__init__.py`. This is the starting, boilerplate code to start a Flask application. Follow your curiousity if you would like to learn more about [`create_app`](https://flask.palletsprojects.com/en/2.0.x/patterns/appfactories/)
+To get started, we will include the following code in `__init__.py`. This is the starting, boilerplate code to start a Flask application. Follow your curiosity if you would like to learn more about [`create_app`](https://flask.palletsprojects.com/en/3.0.x/patterns/appfactories/)
```
kelsey-steven-ada, line 212: It isn't letting me leave a comment on the line because it isn't changed, but we should update the `create_app` function below to remove the `test_config` parameter since we do not add that until pt. 6 with testing.

api-1-setup-read/flask-setup.md
```diff
@@ -67,39 +68,26 @@ These files will list the exact packages and their version numbers to download a
 Here is a sample `requirements.txt` for a Flask project:
```
anselrognlie, line 68: In the previous two paragraphs, can we remove the "will"s? It's unnecessary and complicates the tense.

api-1-setup-read/flask-setup.md
```diff
@@ -231,7 +220,7 @@ def create_app(test_config=None):
     return app
 ```
 
-Configurations to the app can include things like, "Where's the location of our database?," "How do we load different data models, the objects that represent our data?," or "How can we set up template views, called Blueprints?"
+Configurations for the app can include things like, "Where's the location of our database?," "How do we load different data models, the objects that represent our data?," or "How can we set up template views, called Blueprints?"
```
kelsey-steven-ada, line 223: It might be helpful to do a quick run through of the parent ticket for Part 1 to make sure everything was hit, on line 266 below we need to change the instructions for running Flask in debug mode to `flask run --debug`

api-1-setup-read/flask-setup.md
```diff
@@ -148,7 +136,7 @@ System configurations vary, and sometimes the requirements installation step may
 
    ### Common Debugging: Need to `$ pip install --upgrade pip`
```
anselrognlie, line 137: I'd suggest condensing the callout titles. The titls on the one for upgrading `setuptools` now wraps (maybe due to changes in the Learn stylesheet?), leaving the info icon orphaned and looking strange. But if we rewrite that one, we should rewrite the others for consistency.
```suggestion
   ### Common Debugging: Upgrade `pip`
```

(I can't comment on the others since they aren't close enough to a modified region)

api-1-setup-read/flask-setup.md
```diff
@@ -321,7 +310,7 @@ COMMAND   PID  USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
 Python  31756  user    3u  IPv4 0x92416cca3d5db1fb      0t0  TCP localhost:commplex-main (LISTEN)
 ```
 
-We can use the number labelled as `PID` to terminate the other running server with the command `kill -9 31756`, where we would replace `31756` with the `PID` value we got from running `lsof`.
+We can use the number labelled as `PID` (Process Identifier) to terminate the other running server with the command `kill -9 31756`, where we would replace `31756` with the `PID` value we got from running `lsof`.
```
kelsey-steven-ada, line 313: Love adding the full term here for clarity ❤️

api-1-setup-read/flask-setup.md
```diff
@@ -201,14 +189,15 @@ This curriculum will provide a suggested project structure and location:
 │   ├── models
```
anselrognlie, line 189: Another errant "will" in the sentence above this treeview. provide → provides

kelsey-steven-ada (changes requested):
This looks great so far, there are a couple code & instruction updates that we should add plus a fairly trivial suggestion.

api-1-setup-read/flask-setup.md
anselrognlie, line 200: Just pointing out that I'm not _always_ against more future/speculative tenses. While this _could_ be expressed in simple present (The `app/routes` directory is responsible), it also works with "will be" since we're discussing a hypothetical structure.

api-1-setup-read/flask-setup.md
anselrognlie, line 223: In the prior code, we chould leave off the `test_config=None`

anselrognlie: I'm not a fan of the blueprints line
"How can we set up template views, called Blueprints?"

That sounds like it was written before we had a good idea of what all the pieces in Flask were for. It also (at least for a Flask app) doesn't feel very "configuration-y" to me. Really, the only one of the three that sounds like configuration is locating the database, but I guess we're just going for a high-level view of the chunks of code that generally end up in `create_app`.

Anyway, if we want to keep a third thing here, maybe use something like "How do we run different logic in response to different requests?"

api-1-setup-read/flask-setup.md
```diff
@@ -231,7 +220,7 @@ def create_app(test_config=None):
     return app
 ```
 
-Configurations to the app can include things like, "Where's the location of our database?," "How do we load different data models, the objects that represent our data?," or "How can we set up template views, called Blueprints?"
+Configurations for the app can include things like, "Where's the location of our database?," "How do we load different data models, the objects that represent our data?," or "How can we set up template views, called Blueprints?"
 
 ### !callout-danger
```
anselrognlie, line 225: This callout feels very out of place to me. In combination with the general weakness of the description of configuraiton we give above, I feel like this just muddies the waters and creates a potential worry where none is required. Can we cut this?

api-1-setup-read/flask-setup.md
anselrognlie, line 1: For running the server in debug mode, didn't Kelsey recommend a new command for that rather than setting FLASK_ENV?

api-1-setup-read/flask-setup.md
```diff
@@ -321,7 +310,7 @@ COMMAND   PID  USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
 Python  31756  user    3u  IPv4 0x92416cca3d5db1fb      0t0  TCP localhost:commplex-main (LISTEN)
```
anselrognlie, line 310: We may want to update the error message that's shown. Newer Flask versions also detect whether they are on macos and warn about that Apple service that grabs port 5000. We may want to include information here about disabling that service if we haven't included it elsewhere.

api-1-setup-read/flask-setup.md
```diff
@@ -321,7 +310,7 @@ COMMAND   PID  USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
...
 ```
 
-We can use the number labelled as `PID` to terminate the other running server with the command `kill -9 31756`, where we would replace `31756` with the `PID` value we got from running `lsof`.
+We can use the number labelled as `PID` (Process Identifier) to terminate the other running server with the command `kill -9 31756`, where we would replace `31756` with the `PID` value we got from running `lsof`.
 
 </details>
```
anselrognlie, line 316: In the server logs section below, I've had many students get confused about what's going on in the screen shot (they try to type in some of what's actually a part of the log output, and don't see anything happen.

Could we add a Figure caption that this is Flask log out in response to two web requests, and expand the text explanation so the steps to produce something similar are more clear?

anselrognlie (changes requested):
I included some suggestions for a few other areas to look at (couldn't make literal suggestions for most since they weren't in modified areas). We should also make sure code snippets match the updated hello-books, and I think Kelsey had some recommendations to take into account.

api-1-setup-read/flask-setup.md
yangashley: Oh yeah, good point, thanks for the reminder. I'll have a look at the parent ticket in addition to yours and Ansel's review comments.

api-1-setup-read/flask-setup.md
```diff
@@ -23,9 +24,11 @@ We will:
 
 ## Manage Dependencies
 
-Whenever we work on a Python project, we need to consider how we manage our dependencies, the third party packages used by our project.
+As we start writing APIs, we should remember that Flask projects are Python projects. Whenever we work on a Python project, we need to consider how we manage our dependencies, which are the third party packages used by our project.  
 
-`flask` is a package with many of its own dependencies, which can quickly make a mess of our system-wide Python installation!
+Just as we have been doing before, we will want to continue using virtual environments, or `venv`s, to manage our projects.
+
+Since `flask` is a package with many of its own dependencies, we will need to use a `venv` to manage our Flask project so we do not make a mess of our system-wide Python installation.
```
yangashley, line 31: @anselrognlie is this what you had in mind wrt your comment:

"Flask projects are Python projects, just like the other projects we've done.
We use venvs to manage our Python projects.
We will use a venv to manage our Flask project.
Here's a reminder on how to set up a venv for a project when the requirements are supplied to us."

?

anselrognlie: I meant for that flow of topics to replace the entire existing chunk of content from Manage Dependencies to Virtual Environment.

The "Since `flask` is a package" and "In reality, no Python project" paragraphs serve no useful purpose, because we're going to get in the habit of using venvs for everything right away, so dwelling on the fact that it has its own dependencies doesn't inform why we will make a venv.

I'd stay away from phrasing best practices as "we will want to". Stick with we will, or we should. Leave the wanting aspect out of it. It hink this whol section could be boiled down to a single paragraph. Something like:

>As we start writing APIs, we should remember that Flask projects are themselves Python projects. Like any other Python project, we need to manage the packages our project depends on—its dependencies. Like other Python projects, we handle this by creating a `venv` in which to install the packages described in the `requirements.txt` file. Let's briefly review the steps we take to perform this setup.

api-1-setup-read/flask-setup.md
```diff
@@ -21,13 +22,9 @@ We will:
...
 
-## Manage Dependencies
+## Flask Projects are Python Projects
 
-Whenever we work on a Python project, we need to consider how we manage our dependencies, the third party packages used by our project.
-
-`flask` is a package with many of its own dependencies, which can quickly make a mess of our system-wide Python installation!
-
-In reality, no Python project is too small to consider using a virtual environment. But working with a package that brings as many dependencies as does `flask` is reason enough for us to revisit the topic of virtual environments now.
+As we start writing APIs, we should remember that Flask projects are themselves Python projects. Like any other Python project, we need to manage the packages our project depends on—its dependencies. Like other Python projects, we handle this by creating a `venv` in which to install the packages described in the `requirements.txt` file. Let's briefly review the steps we take to perform this setup.
```
yangashley, line 27: @anselrognlie I removed the "Manage Dependencies" section and then incorporated your feedback about discussing Flask projects. 

Does this flow look ok?

anselrognlie: Looks good. I also like the renaming of the section. Nothing special here. Flask projects are just Python projects!

One thing is that now that I read it in context, the "briefly review" here sounds repetative with the "brief review" in the first paragraph of the next section. How about changing this one to "revisit"?

```suggestion
As we start writing APIs, we should remember that Flask projects are themselves Python projects. Like any other Python project, we need to manage the packages our project depends on—its dependencies. Like other Python projects, we handle this by creating a `venv` in which to install the packages described in the `requirements.txt` file. Let's revisit the steps we take to perform this setup.
```

api-1-setup-read/flask-setup.md
```diff
@@ -335,9 +330,11 @@ We can use the server logs to debug our server code. After we run `flask run` fr
 
 These are the logs immediately after a `GET` request to `localhost:5000/i-didnt-define-this-endpoint-in-my-server-code`, which produced a `404` response.
 
-![Screenshot of Terminal, which includes Flask sever startup messages, and output for request/response of a 404 error](../assets/building-an-api/flask-setup_server-logs-404.png)
+![Screenshot of Terminal, which includes Flask server startup messages, and output for request/response of a 404 error](../assets/building-an-api/flask-setup_server-logs-404.png)
```
anselrognlie, line 334: Remove the following blank line and add two spaces after the image to force Learn to treat the caption as inline with the impage, but wrapped to the next line.
```suggestion
![Screenshot of Terminal, which includes Flask server startup messages, and output for request/response of a 404 error](../assets/building-an-api/flask-setup_server-logs-404.png)  
```

api-1-setup-read/flask-setup.md
anselrognlie: Were you going to cut the callout?

api-1-setup-read/flask-setup.md
```diff
@@ -310,7 +293,19 @@ To stop a Flask server:
...
+If something goes wrong when shutting down the server, or if the Terminal where the server was running is somehow closed without stopping the server, the server may not release the address and port where it was running. If that happens, a new server will not be able to run using that address and port, and will fail to start. This may be reported with an error message resembling: 
+
+```bash
+Address already in use
+Port 5000 is in use by another program. Either identify and stop that program, or start the server with a different port.
+On macOS, try disabling the 'AirPlay Receiver' service from System Preferences -> Sharing.
+```
+
+If we know that we have not started the Flask server then we should check that the AirPlay Receiver is disabled like the Terminal output suggests. If it is enabled, we should disable it so that our computer can use port 5000 for running the Flask server.
```
anselrognlie, line 304: Include a bit more detail about what to do.
```suggestion
If we know that we have not already started another Flask server then we should check that the AirPlay Receiver is disabled like the Terminal output suggests. If it is enabled, we should disable it by unchecking its 'On' checkbox, so that our computer can use port 5000 for running the Flask server. 
```

api-1-setup-read/flask-setup.md
```diff
@@ -310,7 +293,19 @@ To stop a Flask server:
...
+```bash
+Address already in use
+Port 5000 is in use by another program. Either identify and stop that program, or start the server with a different port.
+On macOS, try disabling the 'AirPlay Receiver' service from System Preferences -> Sharing.
+```
+
+If we know that we have not started the Flask server then we should check that the AirPlay Receiver is disabled like the Terminal output suggests. If it is enabled, we should disable it so that our computer can use port 5000 for running the Flask server. 
+
+![Screenshot of System Preferences showing the AirPlay Receiver is disabled](../assets/building-an-api/system-preferences-airplay-receiver.png)
+
```
anselrognlie, line 307: Remove the following blank line and add two spaces after the image to improve the caption formatting.
```suggestion
![Screenshot of System Preferences showing the AirPlay Receiver is disabled](../assets/building-an-api/system-preferences-airplay-receiver.png)  
```

api-1-setup-read/flask-setup.md
anselrognlie: Thanks for adding this!

api-1-setup-read/flask-setup.md
```diff
@@ -310,7 +293,19 @@ To stop a Flask server:
...
+_Fig. Screenshot of System Preferences showing the AirPlay Receiver is disabled._
```
anselrognlie, line 308: Give a bit more direction in the caption
```suggestion
_Fig. Screenshot of System Preferences with the `On` checkbox for AirPlay Receiver unchecked, indicating the AirPlay Receiver is disabled._
```

anselrognlie (changes requested):
A few minor suggestions for the added content (thanks for the AirPlay stuff), but wanted to double check on a callout that I suggested cutting.

api-1-setup-read/flask-setup.md
yangashley: ohhh i misread this and cut lines 234-235. Thanks for following up.

anselrognlie (approved):
LGTM!

kelsey-steven-ada (approved):
LGTM!

PR #147, opened 2024-04-12
Building an API pt. 3 - lesson 2: "Models Setup" updates

kelsey-steven-ada:
[Pt. 3 Asana ticket](https://app.asana.com/0/0/1206697443678568/f)
[Pt. 3 Lesson 2 ticket ](https://app.asana.com/0/1205655776549324/1207004723163600/f)


assets/api-3-database-models-read/spiderman_app_book.png
kelsey-steven-ada, line 1: I felt the imports info was really dry so I added this image for some joy, but I'm good removing or replacing it if folks prefer.

api-3-database-models-read/models-setup.md
```diff
@@ -372,25 +523,62 @@ We could tell SQLAlchemy that we would rather use a table name of `books`. There
...
+Arrange the commands below in the correct order to set up a database, create a migration, then apply it.
+
+##### !end-question
+##### !answer
+
+1. `psql -U postgres`
+2. `CREATE DATABASE hello_books_development;`
+3. `flask db init`
+4. `flask db migrate -m "Migration Description"`
+5. `flask db upgrade`
```
anselrognlie, line 542: Should we also include `\q` to mark getting out of `psql`. Otherwise, the implication could be that the `flask` commands are run from `psql`.

anselrognlie: This could also lead to a hint like "think about which commands are run directly in the database, and which are run from the shell."

We might then include an explanation that list the commands in order, and provides a brief summary of each

api-3-database-models-read/models-setup.md
```diff
@@ -372,25 +523,62 @@ We could tell SQLAlchemy that we would rather use a table name of `books`. There
...
+Select all of the options below which could help us troubleshoot if we have an issue connecting to our database
 
 ##### !end-question
+##### !options
+
+* Run the app and try to make a request to a route
+* Check that the database exists locally using `psql`
+* Check that the database connection string doesn’t have typos
+* Check that `app.config['SQLALCHEMY_TRACK_MODIFICATIONS']` is set to False
+* Check that we are setting `app.config['SQLALCHEMY_DATABASE_URI']` in `__init__.py`
```
anselrognlie, line 564: Does the labelled syntax work for checkbox style questions?

api-3-database-models-read/models-setup.md
```diff
@@ -372,25 +523,62 @@ We could tell SQLAlchemy that we would rather use a table name of `books`. There
...
+* title: Models Setup
 ##### !question
 
-What was your biggest takeaway from this lesson? Feel free to answer in 1-2 sentences, draw a picture and describe it, or write a poem, an analogy, or a story.
+Select all of the options below which could help us troubleshoot if we have an issue connecting to our database
 
 ##### !end-question
+##### !options
+
+* Run the app and try to make a request to a route
```
anselrognlie, line 560: I'm on the fence with this one. While sometimes a bad connection can result in a crash before the app starts listening (e.g., didn't set the connection string so it ends up None, or maybe a bad protocol is listed leading to plugin load failure), other configurations aren't detected until a request comes in (bad user/pwd/server/port/dbname). So issuing a request and reading the log error can tell you what the issue is. This overlaps with visually inspecting the connection string, but even the case where the app fails to load can be informative, since it can still help us differentiate between a few cases.

So I guess this feels more like a gotcha to me.

api-3-database-models-read/models-setup.md
```diff
@@ -372,25 +523,62 @@ We could tell SQLAlchemy that we would rather use a table name of `books`. There
...
+* Check that the database exists locally using `psql`
+* Check that the database connection string doesn’t have typos
+* Check that we are setting `app.config['SQLALCHEMY_DATABASE_URI']` in `__init__.py`
 
-My biggest takeaway from this lesson is...
+##### !end-answer
+##### !explanation
 
-##### !end-placeholder
+* Run the app and try to make a request to a route - If the database connection can't be found, often the app will not compile so we cannot use this option.
```
anselrognlie, line 576: I'd go with fails to startup rather than compile. But as mentioned above, I would argue this is a still a useful strategy and should be included as a correct answer, since the error message, whether at startup or when the request is sent, can help us determine the issue.

anselrognlie (changes requested):
There were a lot of changes in this lesson, so there was a lot to chew on. Happy to get together and chat through any of my suggestions or possible alternatives.

api-3-database-models-read/models-setup.md
```diff
@@ -26,17 +27,17 @@ This lesson uses the [Hello Books API](https://github.com/AdaGold/hello-books-ap
 
 ## Removing Hardcoded Data
 
-In a previous lesson we used hardcoded data stored in the variable `BOOKS` to learn how to write routes to _read_ all books, and _read_ one book. This was great for learning, but we were limited to _GET_ routes, as we could not persist any changes to our data given that it was hardcoded.
+In a previous lesson we used hardcoded data stored in the variable `books` to learn how to write routes to _read_ all books, and _read_ one book. This was great for learning, but we were limited to _GET_ routes, as we could not persist any changes to our data given that it was hardcoded.
 
 In this lesson, we will learn how to connect our Flask app to a database. This will allow us to not only write routes to _read_ our data, but also _create_, _update_, and _delete_ it!
 
-Before we go any further, let's comment or remove our hardcoded `BOOKS` data and our two routes from the `app`.
+Before we go any further, let's comment or remove our hardcoded `books` data and our two routes from the `app`.
```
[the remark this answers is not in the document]
anselrognlie: I looked through some other lessons to check for consistency, and it looks like with the rest of the lessons that have a collapsible section after the branch table have them abut, so that should be left as is. The same for any that abut a Piece of Code table. The way everything fits together in those cases make it look like an intentional UI eleemnt.

But for cases of a summary detail following a paragraph, adding the line break helps the paragraph feel less cramped.

api-3-database-models-read/models-setup.md
```diff
@@ -119,32 +129,106 @@ We should get a `CREATE DATABASE` confirmation message from Postgres to indicate
...
+- import the `db` and `migrate` objects we created and call `init_app` on each of them
+
+When we need to tell Flask where to find a database, we do this by providing a _connection string_ that identifies where it is, and how to connect to it.
+
+A connection string resembles the following:
+
+```
+postgresql+psycopg2://postgres:postgres@localhost:5432/REPLACE_THIS_LAST_PART_WITH_DB_NAME
+```
+
```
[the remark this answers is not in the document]
kelsey-steven-ada: Agreed, I can revise that paragraph into a table and it should be easier to take in.

api-3-database-models-read/models-setup.md
```diff
@@ -308,18 +409,51 @@ We can generate database migrations with the following command. This command sho
...
+Example Error Output:
+
+```bash
+(venv) $ flask db migrate -m "adds Book model"
+INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
+INFO  [alembic.runtime.migration] Will assume transactional DDL.
+INFO  [alembic.ddl.postgresql] Detected sequence named 'book_id_seq' as owned by integer column 'book(id)', assuming SERIAL and omitting
+INFO  [alembic.env] No changes in schema detected.
+```
+
```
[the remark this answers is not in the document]
kelsey-steven-ada: What do you think about this updated wording & callout?

```suggestion
If no change is detected, we should reflect on whether any recent model changes would require the database to be updated. For instance, if we add a helper method to a model class, only our code is affected, it doesn't change the data stored in the database. In that case, generating a migration isn't necessary, and if we try to do so, no changes will be detected. 

However, if we added a new attribute to our model, that _does_ need to update the database. We need to generate a migration to tell the database to create a new column for the new attribute. If generating a migration in that case detects no changes, we'll need to do more research to determine the issue.
	
### !callout-info

## Database Migration Issues 

Migrations can be a bit finicky to keep in sync, especially when multiple people are working on a project. We need to be careful about coordinating who is creating migrations and ensuring they are shared with the team to keep everyone's development database in sync. 

<br/>

We won't have the opportunity to invest much time in troubleshooting migrations during this series, but a common resolution can be to clear out the database and generate a fresh set of migrations. More details on how to do this are described in the "Resolving Schema Conflicts" reading a bit later on.

### !end-callout
```

api-3-database-models-read/models-setup.md
kelsey-steven-ada: Ahh, that's fair and something to avoid. I may just remove this option.

assets/api-3-database-models-read/spiderman_app_book.png
kelsey-steven-ada: I made another version that has `Alembic` instead of `app.py`, I'm happy to sub that one in ^_^

anselrognlie: I still wish I could remember where the other copy was. Did we maybe remove it at some point? We don't want to be called out for having stale memes! 😆

api-3-database-models-read/models-setup.md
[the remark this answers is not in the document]
anselrognlie: Yep. That's it.

api-3-database-models-read/models-setup.md
kelsey-steven-ada: I will play around with it and find out, I'm not sure

api-3-database-models-read/models-setup.md
```diff
@@ -119,32 +129,106 @@ We should get a `CREATE DATABASE` confirmation message from Postgres to indicate
```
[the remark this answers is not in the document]
anselrognlie: Looks good. For the heading, how about

```
## The `psql` Prompt Provides Input Feedback
```

api-3-database-models-read/models-setup.md
```diff
@@ -251,40 +324,52 @@ In model definitions, if we don't like the default name that SQLAlchemy picks we
 
 ### Making the `Book` Model Visible to the `app` in `__init__.py`
```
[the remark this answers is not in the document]
anselrognlie: I'm good with going all the way to Alembic. In the past, we've sort of downplayed its involvement, but at the end of the day, all those `flask db` commands are Alembic commands, and those are the commands that need the models to be visible prior to explicitly using them anywhere else.

api-3-database-models-read/models-setup.md
```diff
@@ -300,6 +385,22 @@ Once we have created our database and configured the connection string, we can d
...
+  Generating /Users/user_name/Documents/curriculum_changes/hello-books-api/migrations/alembic.ini ...  done
+  Please edit configuration/connection/logging settings in
+  '/Users/user_name/Documents/hello-books-api/migrations/alembic.ini' before proceeding.
+```
+
+The final message in the output `"Please edit configuration..."` is just a recommendation. We will not be discussing logging data about our API as part of this series, so we do not need to take any action with the listed file. Once we see this message we are ready to keep moving forward!
+
 ### Generate Migrations After Each Model Change
 
 We can generate database migrations with the following command. This command should be run every time there's a change to a file in the `models` folder.
```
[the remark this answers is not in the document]
anselrognlie: Works for me!

api-3-database-models-read/models-setup.md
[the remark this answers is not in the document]
anselrognlie: Love this, especially mentioning the effect of having multiple contributors. That was something that had crossed my mind that we should be more explicit about at least in roundtables, but seemed like it might be too much to give the consideration it needs what with everything else we're doing. Like it feels like there's enough content to justify a lesson on coordinating migrations on projects with multiple contributors.

But having any kind of mention is better than nothing! 🙏

api-3-database-models-read/models-setup.md
```diff
@@ -119,32 +129,106 @@ We should get a `CREATE DATABASE` confirmation message from Postgres to indicate
...
+There are a couple steps to creating the database connection. We will:
+1. create objects that represent our database and migration tools
+2. register these objects with our application and tell Flask where find our new database in `create_app`
 
-A connection string resembles the following:
+### Set Up the Database and Migrations
 
+To set up our connection with SQLAlchemy, we need to create an instance of SQLAlchemy and give it information on how we want to create our database backed models. We do this by passing the SQLAlchemy constructor a class name to use when generating our model classes.
+
+That might sound like a lot of work, but SQLAlchemy provides a class named `DeclarativeBase` that provides all of the operations our database backed models will need. We will make our own `Base` model class that is a subclass of `DeclarativeBase`.
```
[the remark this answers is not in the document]
anselrognlie: I agree with swapping "that" to "which", since it's a non-restrictive clause. In that case, a comma is also recommended. In the second sentence, "that will be" has closer verb agreement with "we will make".
```suggestion
That might sound like a lot of work, but SQLAlchemy provides a class named `DeclarativeBase`, which provides all of the operations our database backed models will need. We will make our own `Base` model class that will be a subclass of `DeclarativeBase`. 
```

api-3-database-models-read/models-setup.md
```diff
@@ -119,32 +129,106 @@ We should get a `CREATE DATABASE` confirmation message from Postgres to indicate
...
-The text `REPLACE_THIS_LAST_PART_WITH_DB_NAME` should be replaced with our database's name, `hello_books_development`.
+With our `Base` model in hand, we can create a new file in the `app` folder `db.py` which will manage creating our database and migration objects:
 
-### Configure the Database in `app/__init__.py`
+```bash
+$ touch app/db.py
+```
 
-We can examine this code, which configures the database to use SQLAlchemy appropriately for our app.
+We can read through the contents of that file:
```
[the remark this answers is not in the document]
anselrognlie: That works, but I'm still bothered by "we can".

```suggestion
Let's look over the contents for `db.py` before copying it to our local file:
```

assets/api-3-database-models-read/spiderman_app_book.png
kelsey-steven-ada: No crusty memes here! We might have removed it, I did a quick check and don't see it in Unit 1 or other Unit 2 lessons at least,  but I haven't looked at Industry prep/ Unit 3.

api-3-database-models-read/models-setup.md
```diff
@@ -372,25 +543,79 @@ We could tell SQLAlchemy that we would rather use a table name of `books`. There
...
+b| Check that the database connection string doesn’t have typos
+c| Check that `app.config['SQLALCHEMY_TRACK_MODIFICATIONS']` is set to False
+d| Check that we are setting `app.config['SQLALCHEMY_DATABASE_URI']` in `__init__.py`
+
+##### !end-options
+##### !answer
+
+a| Check that the database exists locally using `psql`
+b| Check that the database connection string doesn’t have typos
+d| Check that we are setting `app.config['SQLALCHEMY_DATABASE_URI']` in `__init__.py`
```
anselrognlie, line 606: With the labeled options, the solution should be able to be listed as just
```suggestion
a|
b|
d|
```

api-3-database-models-read/models-setup.md
```diff
@@ -372,25 +543,79 @@ We could tell SQLAlchemy that we would rather use a table name of `books`. There
...
+a| Check that the database exists locally using `psql` - We should check if the database exists and create one if it is not present.
 
-##### !placeholder
+b| Check that the database connection string doesn’t have typos - If we misspell any part of the connection string, we will not be able to connect to the database.
 
-My biggest takeaway from this lesson is...
+c| Check that app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] is set to False - This setting doesn't affect our ability to reach the database
 
-##### !end-placeholder
+d| Check that we are setting app.config['SQLALCHEMY_DATABASE_URI'] in `__init__.py` - We need to give this dictionary key the value of our database connection string for Flask to know where the database is.
```
anselrognlie, line 617: In the explanation, the labels aren't treated specially, so a plain bullet will have cleaner formatting.

api-3-database-models-read/models-setup.md
```diff
@@ -372,25 +543,79 @@ We could tell SQLAlchemy that we would rather use a table name of `books`. There
...
+##### !hint
+
+Think about which commands are run directly in the database, and which are run from the shell.
+
+##### !end-hint
+##### !explanation
+
+1. `psql -U postgres` - start the `psql` command line interface with the user postgres
+2. `CREATE DATABASE hello_books_development;` - Create a database named `hello_books_development`
+3. `\q``- Quit the `psql` command line interface
```
anselrognlie, line 575:
```suggestion
3. `\q` - Quit the `psql` command line interface
```

api-3-database-models-read/models-setup.md
```diff
@@ -117,34 +131,121 @@ CREATE DATABASE hello_books_development;
...
+There are a couple steps to creating the database connection. We will:
+1. initialize objects to represent our database, and migration tools
+2. register these objects with our application and tell Flask where to find our new database in `create_app`
 
-A connection string resembles the following:
+### Set Up the Database and Migrations
 
+To set up our connection with SQLAlchemy, we need to create an instance of `SQLAlchemy` and give it information on how we want to create our database backed models. We do this by passing the SQLAlchemy constructor a class name to use when generating our model classes.
+
+That might sound like a lot of work, but SQLAlchemy provides a class named `DeclarativeBase` which provides all of the operations our database backed models will need. We will make our own `Base` model class that will be a subclass of `DeclarativeBase`.
```
anselrognlie, line 154: I'd still include this comma for the non-restrictive clause
```suggestion
That might sound like a lot of work, but SQLAlchemy provides a class named `DeclarativeBase`, which provides all of the operations our database backed models will need. We will make our own `Base` model class that will be a subclass of `DeclarativeBase`. 
```

api-3-database-models-read/models-setup.md
```diff
@@ -117,34 +131,121 @@ CREATE DATABASE hello_books_development;
...
+    pass
 ```
 
-This tells Flask to connect to our database using the `psycopg2` package we installed from our `requirements.txt`. It connects using the `postgres` protocol using the `postgres` user on the local machine running at port `5432`.
+Our code here is short, with our only actions being to 
+1. import the parent class `DeclarativeBase`, and
+2. create an empty subclass `Base` that we can give to our SQLAlchemy constructor.
 
-The text `REPLACE_THIS_LAST_PART_WITH_DB_NAME` should be replaced with our database's name, `hello_books_development`.
+With our `Base` model in hand, we can create a new file in the `app` folder `db.py` which will manage creating our database and migration objects:
```
anselrognlie, line 177: non-restrictive clause comma
```suggestion
With our `Base` model in hand, we can create a new file in the `app` folder `db.py`, which will manage creating our database and migration objects:
```

api-3-database-models-read/models-setup.md
```diff
@@ -117,34 +131,121 @@ CREATE DATABASE hello_books_development;
...
+db = SQLAlchemy(model_class=Base)
 migrate = Migrate()
+```
+
+Then we can update our `db.py` to contain the same code.
+
+The file above does a few things:
+
+1. First we import `SQLAlchemy`, `Migrate`, and our newly created `Base` class
+2. Next, we create an instance of SQLAlchemy that we will call `db` and pass it our `Base` class as the constructor argument. We will use the `db` object when we need to interact with the database to perform operations like creating or updating records.
```
anselrognlie, line 199: non-restrictive clause
```suggestion
2. Next, we create an instance of SQLAlchemy, which we will call `db`, and pass it our `Base` class as the constructor argument. We will use the `db` object when we need to interact with the database to perform operations like creating or updating records.
```

api-3-database-models-read/models-setup.md
```diff
@@ -117,34 +131,121 @@ CREATE DATABASE hello_books_development;
...
-def create_app(test_config=None):
+The required format of a connection string can vary depending on the library being used to communicate with the database. The packages we are using (SQLAlchemy and psycopg2) structure their connection strings to look like a URL. Consider the example connection string shown here:
+
+```
+postgresql+psycopg2://postgres:postgres@localhost:5432/REPLACE_THIS_LAST_PART_WITH_DB_NAME
+```
+
+| <div style="min-width:250px;"> Piece of Code </div> | Notes|
+| --------------------------------------------------- | ---- |
+| `postgresql+psycopg2` | This tells Flask to connect to our database using the `psycopg2` package we installed from our `requirements.txt`. |
```
anselrognlie, line 236: Thanks for including the table! How about chunking them up like this (and the following)
```suggestion
| `postgresql+psycopg2` | Contains information about the protocols to use for connecting to the database. This tells Flask to connect to our `postgresql` database using the `psycopg2` package we installed from our `requirements.txt`. |
```

api-3-database-models-read/models-setup.md
```diff
@@ -117,34 +131,121 @@ CREATE DATABASE hello_books_development;
...
+| `://postgres:postgres` | We connect to the database using the `postgres` protocol and the user `postgres` |
```
anselrognlie, line 237:
```suggestion
| `://` | Required URL syntax to separate the protocol part from the user or host part. |
| `postgres:postgres` | Contains information about the user credentials to use to connect to a resource, here, the database. We connect to the database using the `postgres` user (the part before the `:`), using the password `postgres` (the part after the `:`). We never actually set a database password, but when connecting to a local PostgreSQL service, the supplied password value is ignored anyway and could be omitted. It is included here for completeness. |
```

api-3-database-models-read/models-setup.md
```diff
@@ -117,34 +131,121 @@ CREATE DATABASE hello_books_development;
...
+| `@localhost:5432` | We are connecting to the local machine (`localhost`) running at port `5432` |
```
anselrognlie, line 238:
```suggestion
| `@` | Required URL syntax to separate the user part from the host part. |
| `localhost:5432` | Contains information about the host (the machine "hosting" the database) location. We are connecting to the local machine (`localhost`) running at port `5432` |
```

api-3-database-models-read/models-setup.md
```diff
@@ -117,34 +131,121 @@ CREATE DATABASE hello_books_development;
...
+| `REPLACE_THIS_LAST_PART_WITH_DB_NAME` | This text should be replaced with the name of the database we're connecting to. For the hello-books-api project, we will replace it with our development database's name, `hello_books_development`. |
```
anselrognlie, line 239:
```suggestion
| `/` | Required URL syntax to separate the host part from the path part. |
| `REPLACE_THIS_LAST_PART_WITH_DB_NAME` | Normally contains path information about the resource being accessed. This text should be replaced with the name of the database we're connecting to. For the hello-books-api project, we will replace it with our development database's name, `hello_books_development`. |
```

api-3-database-models-read/models-setup.md
```diff
@@ -117,34 +131,121 @@ CREATE DATABASE hello_books_development;
...
+To finish configuring the database, we will make 2 changes to `create_app`:
+- Supply a _connection string_ to tell Flask where to find our new database
+- import the `db` and `migrate` objects we created and use their `init_app` methods to register them with our `app`
+
+When we need to tell Flask where to find a database, we do this by providing a _connection string_ that identifies where it is, and how to connect to it.
 
-def create_app(test_config=None):
+The required format of a connection string can vary depending on the library being used to communicate with the database. The packages we are using (SQLAlchemy and psycopg2) structure their connection strings to look like a URL. Consider the example connection string shown here:
+
+```
```
anselrognlie, line 230: undecorated, there's some weird coloration going on
```suggestion
```text
```

api-3-database-models-read/models-setup.md
```diff
@@ -163,73 +264,72 @@ Then, we can edit our `app/__init__.py` file so it looks identical.
...
-1. Connect the model seamlessly with our SQL database, mapping attributes to table columns
-1. Give our Flask code the ability to work with `Book` instances, taking advantage of OOP
-1. Enable us to access instances of `Book`, where each instance corresponds to a row in our database
+SQLAlchemy has changed over the years, both in syntax and in meaning. The current syntax borrows from Python's type hinting system to make the column declarations more concise, as well as to make it easier for tools to check our code for errors. Additionally, some behaviors that used to be the default must now be set explicitly.
 
-```python
-from app import db
+<br />
 
+Recall that SQL permits us to define columns that allow NULL values. In the past, SQLAlchemy would default to allowing NULL values for all columns unless otherwise marked. In the current version, columns using `Mapped` types as above, that are not explicitly marked `Optional`, are considered non-nullable; they are required attributes when creating a new instance of a model.
```
anselrognlie, line 303: shifts the aside to avoid splitting the restrictive clause
```suggestion
Recall that SQL permits us to define columns that allow NULL values. In the past, SQLAlchemy would default to allowing NULL values for all columns unless otherwise marked. In the current version, columns using `Mapped` types that are not explicitly marked `Optional`, as above, are considered non-nullable; they are required attributes when creating a new instance of a model.
```

api-3-database-models-read/models-setup.md
```diff
@@ -300,26 +385,95 @@ Once we have created our database and configured the connection string, we can d
...
+  Creating directory '/Users/user_name/Documents/hello-books-api/migrations/versions' ...  done
+  Generating /Users/user_name/Documents/hello-books-api/migrations/script.py.mako ...  done
+  Generating /Users/user_name/Documents/hello-books-api/migrations/env.py ...  done
+  Generating /Users/user_name/Documents/curriculum_changes/hello-books-api/migrations/README ...  done
+  Generating /Users/user_name/Documents/curriculum_changes/hello-books-api/migrations/alembic.ini ...  done
+  Please edit configuration/connection/logging settings in
+  '/Users/user_name/Documents/hello-books-api/migrations/alembic.ini' before proceeding.
+```
+
+Although the command `flask db init` sounds like it should initialize our database, in reality, it doesn't look at the database at all! This command creates our project's migrations folder, which Alembic will use to store its generated migration files.
```
anselrognlie, line 402:
```suggestion
Although the command `flask db init` sounds like it should initialize our database, in reality, it doesn't look at the database at all! This command creates our project's `migrations` folder, which Alembic will use to store its generated migration files.
```

api-3-database-models-read/models-setup.md
```diff
@@ -300,26 +385,95 @@ Once we have created our database and configured the connection string, we can d
...
+
+```bash
+(venv) $ flask db migrate -m "adds Book model"
+INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
+INFO  [alembic.runtime.migration] Will assume transactional DDL.
+INFO  [alembic.ddl.postgresql] Detected sequence named 'book_id_seq' as owned by integer column 'book(id)', assuming SERIAL and omitting
+INFO  [alembic.env] No changes in schema detected.
+```
+
+If no change is detected, we should reflect on whether any recent model changes would require the database to be updated. For instance, if we add a helper method to a model class, only our code is affected, it doesn't change the data stored in the database. In that case, generating a migration isn't necessary, and if we try to do so, no changes will be detected.
```
anselrognlie, line 445: separates the two clauses. could use ; instead
```suggestion
If no change is detected, we should reflect on whether any recent model changes would require the database to be updated. For instance, if we add a helper method to a model class, only our code is affected. It doesn't change the data stored in the database. In that case, generating a migration isn't necessary, and if we try to do so, no changes will be detected. 
```

api-3-database-models-read/models-setup.md
```diff
@@ -334,12 +488,25 @@ At the introductory level of Flask, it's unlikely that we will touch many of the
```
anselrognlie, line 488: I had made a suggestion to replace the "At the introductory..." passage above with
```md
In situations where we need to have more control over the migration process, we may need to review a generated migration file and modify the code in either the `upgrade` or `downgrade` functions. Pregenerated comments in those functions even encourage us to make adjustments! We probably won't need to make such modifications at this point, but it's good practice to at least review the migrations before applying them in order to confirm that the generated result matches our expectations.

Skimming through the generated migrations is good practice for reading code and making assumptions about what it will do based on the context.
```

anselrognlie (changes requested):
Thanks for the connection string table! I have some suggestions there and for the CFU. There was also a paragraph update I suggested but I don't see, and then assorted minor updates.

api-3-database-models-read/models-setup.md
kelsey-steven-ada: Definitely more complete information, thanks for including the suggestions!

api-3-database-models-read/models-setup.md
```diff
@@ -330,16 +487,31 @@ A neat side-effect about generating migrations is that we get to appreciate the
 └── requirements.txt
 ```
 
-At the introductory level of Flask, it's unlikely that we will touch many of the files in the `migrations` folder.
+In situations where we need more control over the migration process, we may need to review a generated migration file and modify either the `upgrade` or `downgrade` functions. Pregenerated comments in those functions even encourage us to make adjustments! 
+
+We probably won't need to make such modifications at this point, but it's a good habit to at least review the migrations before applying them in order to confirm that the generated result matches our expectations. Skimming through the migration files is also good practice for reading code and making assumptions about what it will do based on the context.
```
kelsey-steven-ada, line 492: I was struggling with the one sentence alone that didn't feel fully thematically linked and made some small adjustments, let me know what you think.

anselrognlie: Works for me!

anselrognlie (approved):
LGTM. Ship it!

PR #148, opened 2024-04-15
Updates for API Lesson 1 Part 5

yangashley:
Asana task: https://app.asana.com/0/1205655776549324/1207004723163591/f


api-1-setup-read/read-all-books.md
```diff
@@ -128,72 +151,52 @@ Again, these lines make it so that our `Blueprint` is recognized by our Flask `a
...
-| `jsonify(books_response)`                           | `books_response` contains a list of book dictionaries. To turn it into a `Response` object, we pass it into `jsonify()`. This will be our common practice when returning a list of something. When we are returning strings or dictionaries, we can use `make_response`, which we will learn about later. |
-| `return ...`                                        | We must return our response. By default, a response with no specified status code returns `200 OK`                                                                                                                                                       |
+| `return books_response`                                        | We must return our response. By default, a response with no specified status code returns `200 OK`                                                                                                                                                       |
+
+## What Kind of Data Should We Return to the Client?
```
yangashley, line 191: I could use some feedback here. In addition to the wording, let me know if you think I need to add additional content here.

api-1-setup-read/read-all-books.md
```diff
@@ -44,81 +55,93 @@ class Book:
...
+| `books_bp = Blueprint("books_bp", __name__, url_prefix="/books")`        | Our `Blueprint` instance. We'll use it to group routes that start with `/books`. `"books"` is the debugging name for this `Blueprint`. `__name__` provides information the blueprint uses for certain aspects of routing. We should use this blueprint for all of our RESTful routes that start with `/books`! 
 
 ### Registering a Blueprint
 
 Now that we have defined our `books_bp` blueprint, Flask requires us to "register the blueprint" with our `app`.
 
-Let's return to the code in `app/__init__.py`. Inside our `create_app` let's add two lines to register our `books_bp` Blueprint inside the `create_app` function:
+Let's return to the code in `app/__init__.py`. From our previous lesson "Hello World Routes", we already have `Flask` imported. Now we need to import our `books_bp` blueprint and register it inside our `create_app` function, as shown below.
```
kelsey-steven-ada, line 133: To reduce dependencies between lessons in case content moves in the future, we want to avoid mentioning specific lessons unless it's necessary.

```suggestion
Let's return to the code in `app/__init__.py`. Taking a look at our imports, we brought in `Flask` in a previous lesson, and now we need to import our new `books_bp` blueprint. Next, we register `books_bp` inside the `create_app` function, as shown below.
```

api-1-setup-read/read-all-books.md
```diff
@@ -128,72 +151,52 @@ Again, these lines make it so that our `Blueprint` is recognized by our Flask `a
...
+| `from app.models.book import books`               | An import statement that imports the list `books` from a module called `book` in the `app/models` directory. Recall that Python modules are files with the `.py` extension so the `book` module references the `book.py` file.                                                                                                                             |
+| `@books_bp.get("")`               | A decorator that uses the `books_bp` Blueprint to define an endpoint and the accepted HTTP method. The following function will execute whenever a matching HTTP request is received.                                                                                                                            |
+| `def get_all_books():`                                 | This function will execute whenever a request that matches the decorator is received. The name of this function doesn't affect how requests are routed to this method. Common choices for a function name could include matching the route path, or using any other good, descriptive Python function name. |
```
kelsey-steven-ada, line 184: I'm trying to update places we use terms like "good" that could be more descriptive when I see them in the curriculum.

```suggestion
| `def get_all_books():`                                 | This function will execute whenever a request that matches the decorator is received. The name of this function doesn't affect how requests are routed to this method. Common choices for a function name could include matching the route path, or using any other clear, descriptive Python function name. |
```

api-1-setup-read/read-all-books.md
```diff
@@ -128,72 +151,52 @@ Again, these lines make it so that our `Blueprint` is recognized by our Flask `a
...
+| `books_response = []`                                       | We create an empty list called `books_response` that will eventually hold book dictionaries and be returned to the client that made the request. |
+| `for book in books:`                                | We iterate over each `book` in `books` so we can collect their data and format it into a response                                                                                                                                                          |
+| `books_response.append( ... )`                      | We append book dictionaries to the list                                                                                                                                                                                          |
```
kelsey-steven-ada, line 187:
```suggestion
| `books_response.append( ... )`                      | We append book dictionaries to the list to build up our response                                                                                                                                                                                         |
```

api-1-setup-read/read-all-books.md
```diff
@@ -128,72 +151,52 @@ Again, these lines make it so that our `Blueprint` is recognized by our Flask `a
...
-| `jsonify(books_response)`                           | `books_response` contains a list of book dictionaries. To turn it into a `Response` object, we pass it into `jsonify()`. This will be our common practice when returning a list of something. When we are returning strings or dictionaries, we can use `make_response`, which we will learn about later. |
-| `return ...`                                        | We must return our response. By default, a response with no specified status code returns `200 OK`                                                                                                                                                       |
+| `return books_response`                                        | We must return our response. By default, a response with no specified status code returns `200 OK`                                                                                                                                                       |
```
kelsey-steven-ada, line 189:
```suggestion
| `return books_response`                                        | We must return a response from our route functions. By default, a response with no specified status code returns `200 OK`                                                                                                                                                       |
```

api-1-setup-read/read-all-books.md
```diff
@@ -128,72 +151,52 @@ Again, these lines make it so that our `Blueprint` is recognized by our Flask `a
...
+
+## What Kind of Data Should We Return to the Client?
+
+What kind of data should route functions return? A list? A dictionary? It depends.
+
+For the `get_all_books` route, we returned a list of book dictionaries representing book data from each of the books in our "database". There will be times when a client will send a request for data for a single book so we would only need to return a single book dictionary instead of a list of book dictionaries.
```
kelsey-steven-ada, line 195:
```suggestion
For the `get_all_books` route, we returned a list of book dictionaries representing data from each of the books in our "database". In the future we'll see an example of a client sending a request for a single book's data; in that case it would be appropriate to return a single book dictionary instead of a list of dictionaries. 
```

api-1-setup-read/read-all-books.md
```diff
@@ -128,72 +151,52 @@ Again, these lines make it so that our `Blueprint` is recognized by our Flask `a
...
+| `return books_response`                                        | We must return our response. By default, a response with no specified status code returns `200 OK`                                                                                                                                                       |
+
+## What Kind of Data Should We Return to the Client?
+
+What kind of data should route functions return? A list? A dictionary? It depends.
+
+For the `get_all_books` route, we returned a list of book dictionaries representing book data from each of the books in our "database". There will be times when a client will send a request for data for a single book so we would only need to return a single book dictionary instead of a list of book dictionaries. 
+
+It's easy to work with lists and dictionaries because Flask conveniently converts these two data types into JSON responses when they are returned from a route. Generally speaking, the APIs we'll be writing here at Ada will accept requests and returns responses with JSON as the format for the request body and response body, respectively.
```
kelsey-steven-ada, line 197: I think it could be helpful to talk a little about being consistent with what our APIs send back here.

```suggestion
When building an API, it's considered a best practice to be consistent about what kind of data a user can expect our endpoints to return. This makes it easier for users to know what to expect, and users can often write less code because they don't need to detect and handle multiple formats from the response. 

Generally speaking, the APIs we write through this curriculum will accept requests and return responses with JSON as the format for the request body and response body, respectively. Flask conveniently converts list and dictionary data types into JSON responses when they are returned from a route, so we will be focusing on using these data types for our responses.
```

api-1-setup-read/read-all-books.md
```diff
@@ -128,72 +151,52 @@ Again, these lines make it so that our `Blueprint` is recognized by our Flask `a
...
+
+What kind of data should route functions return? A list? A dictionary? It depends.
+
+For the `get_all_books` route, we returned a list of book dictionaries representing book data from each of the books in our "database". There will be times when a client will send a request for data for a single book so we would only need to return a single book dictionary instead of a list of book dictionaries. 
+
+It's easy to work with lists and dictionaries because Flask conveniently converts these two data types into JSON responses when they are returned from a route. Generally speaking, the APIs we'll be writing here at Ada will accept requests and returns responses with JSON as the format for the request body and response body, respectively. 
+
+If we return other objects, Flask might not be able to convert the object into a response body with JSON as the format or it might incorrectly make assumptions about the format it should use to send a response body back to the client. Therefore, we should convert an instance of an object into a dictionary which we can return on its own or append to a list of objects that have been converted into dictionaries to return as a response.
```
kelsey-steven-ada, line 199:
```suggestion
If we return other objects, Flask may not be able to convert the object into a JSON formatted response body, or it might make incorrect assumptions about which format to use when sending a response body back to the client. Therefore, we should convert an instance of an object into a dictionary which we can return on its own–or append to a list of objects that have been converted into dictionaries to return–as a response. 
```

api-1-setup-read/read-all-books.md
```diff
@@ -234,33 +238,52 @@ Remember to use all debugging tools:
 
 ### !end-callout
 
+## Check for Understanding
```
kelsey-steven-ada, line 241: I can't leave a message on the section since it wasn't changed, but I'm wondering if it's useful to add something about "checking the Flask Server logs in Terminal" to the debugging tools listed above? I want to make sure folks are comfortable looking at the server output for errors (and when we get to it in later lessons, using psql to confirm database contents)

api-1-setup-read/read-all-books.md
```diff
@@ -234,33 +238,52 @@ Remember to use all debugging tools:
...
+
+<!-- Question 1 -->
 <!-- prettier-ignore-start -->
 ### !challenge
-* type: tasklist
-* id: f6e59db1-edbc-40e1-af4e-f8b326899df9
-* title: Create and Read, Getting All Books Endpoint
+* type: ordering
+* id: 7GgQrh
+* title: Reading All Books
```
kelsey-steven-ada, line 248: I believe we're using the name of the lesson as the title for all in-lesson check for understanding questions. This feedback applies to the second Check for Understanding question as well.

```suggestion
* title: Read All Books Endpoint
```

kelsey-steven-ada (commented):
Looking really good, just some minor suggestions & questions for discussion

api-1-setup-read/read-all-books.md
```diff
@@ -16,26 +17,36 @@ In this lesson we will build our Hello Books API to fulfill the following featur
...
+
+In our previous lesson "Hello World Routes" we created our first blueprint and wrote our first routes. Now that we have had a chance to get familiar with Flask, we can clean up the work we just did on our local `main` branch since we will not need it for the rest of our Hello Books API. 
+
+We can reference the changes on the `01c-hello-world` branch from the GitHub web interface to see what we need to clean up. We should either delete or comment out the following pieces of code to prepare for writing our first Hello Books API route:
+
+- Line 2 in `app/__init__.py`: The import statement that imports the `hello_world_bp` blueprint 
+- Line 8 in `app/__init__.py`: Remove the code that registers `hello_world_bp` with `app`
+- The entire file `app/routes/hello_world_routes.py`
+
 ## Getting All Books Endpoint: Preparation
```
kelsey-steven-ada, line 30: Even I'm a little overwhelmed by the blocks of text in this lesson series - this isn't necessary, but I'm thinking that dropping in some kind of meme to break up the text and dryness as students jump in might be helpful. What do you think?

Some examples of the kinds of thing I was thinking:
- https://media.makeameme.org/created/the-data-all.jpg
- https://crawlbase.com/blog/best-data-memes/data-scraping-meme.jpg

yangashley: Ha yes, I know what you mean. Good call about breaking it up with an image. I'll add one in

yangashley: How do you feel about the flow when I add the meme here:
image

api-1-setup-read/read-all-books.md
```diff
@@ -206,7 +213,7 @@ Be sure to specify that this endpoint is `""`. When combined with the `book_bp`'
 <details>
 
 <summary>Is there a way to declare "/" as a valid route path?</summary>
-
+<br/>
 For a little more flexibility, we _could_ choose to use `"/"` as the route path and include the keyword argument `strict_slashes=False`. This tells the route to treat a URI the same whether or not it ends in `/`. Accepting either variation can make using our API a little easier for our clients.
```
kelsey-steven-ada, line 217: In my updates for pt. 3 lesson 3 I realized this content about blueprints & "/" is fully duplicated in that lesson. Rather than having it in both places, could we update this to include the changes I was going to make for clarity and I'll remove it from pt. 3 entirely?

```suggestion
For a little more flexibility, we _could_ choose to use `"/"` as the route path and include the keyword argument `strict_slashes=False` in our route decorator:
```python
@books_bp.get("", strict_slashes=False)
```
 This tells the route to treat a URI the same whether or not it ends in `/`. Accepting either variation can make using our API a little easier for clients.
```

kelsey-steven-ada (approved):
Approved with a question that came up while I was working on pt. 3 lesson 3

api-1-setup-read/read-all-books.md
yangashley: Yeah, I agree with you that we can cut it from Part 3 and I'll update the callout with your suggestion.

PR #155, opened 2024-04-30
Flask, SqlAlchemy updates for api part 4 lesson 1, query params

anselrognlie:


api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,542 @@ From this example, we can observe:
...
-    for book in books:
-        books_response.append({
-            "id": book.id,
-            "title": book.title,
-            "description": book.description
-        })
+    # remaining code as before
+```
+
+We check for `title` in the query params. If it's not present, the `get()` call will return `None`, a falsy value. We check whether `title_param` is truthy to decide which branch of the conditional to take. Truthy indicates we got a value for the `title` query param, so we should filter by it. Falsy indicates we did not get a value, so we should get all books as before.
```
kelsey-steven-ada, line 282: Not a necessary change, but I'm wondering if this section is a little easier to read as a list.

```suggestion
In the code above our changes:
1. check for `title` in the query params. If it's not present, the `get()` call will return `None`, a falsy value. 
2. check whether `title_param` is truthy to decide which branch of the conditional to take. Truthy indicates we got a value for the `title` query param, so we should filter by it. Falsy indicates we did not get a value, so we should get all books as before.
```

api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,542 @@ From this example, we can observe:
...
+As before, this request takes no request body.
+
+Let's add some descriptions to our previous sample books.
+
+<!-- prettier-ignore-start -->
+| `title`  | `description` |
+| - | - |
+| `10,000 Apples` | `When all you need is an orange... Now isn't that ironic? Don't you think?` |
+| `It's Not Easy Being an Orange` | `Limes and apples aren't the only ones with problems, you know!` |
+| `An Apple a Day` | `When Shay fell for a doctor, would a love of apples spell disaster for their happiness?` |
```
kelsey-steven-ada, line 371: 😆

api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,542 @@ From this example, we can observe:
...
 ## Pro-tip: Check `request` Spelling
 
-The popular package `requests`, used for making HTTP requests, is not the same as the `request` that represents the HTTP request. If we are working on a web API project that also calls external APIs by using the `requests` package, we should be especially careful of whether we have written `request` or `requests`!
+The popular package `requests`, used for making HTTP requests, is not the same as the `request` that represents the HTTP request. If we are working on a web API project that also calls external APIs by using the `requests` package, we should be especially careful of whether we have written `request` or `requests`. It's very easy to accept the wrong one from an autosuggest!
 
 ### !end-callout
 
-## Using Query Params
-
-The way we choose to use query params will depend on our project needs.
```
kelsey-steven-ada, line 128: So happy to have this removed ^_^

api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,542 @@ From this example, we can observe:
...
+
+If we know what we're trying to do, but not how, we can try searching for the answer. We can try searching for something like `SQLAlchemy filter by partial string` and see what we find. Among the results, we might learn that we can use the [`like()`](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.ColumnElement.like) method to filter by a partial string. We might also encounter results describing the [`contains()`](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.ColumnElement.contains) method, which works similarly. We'll stick with `like()`, because it resembles a standard operator in SQL called [`LIKE`](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE), which is commonly used for this purpose.
+
+SQLAlchemy assumes that developers using it have some familiarity with SQL, so it provides methods that are similar to SQL operators. The more we practice SQL in general, the better we'll be able to ask questions about how to use SQLAlchemy.
+
+IN SQL, the query we're trying to write would resemble:
+
+```sql
+SELECT * FROM book WHERE title LIKE '%apple%' ORDER BY id;
```
kelsey-steven-ada, line 300: Love that we're tying this back to SQL more to remind folks of those skills and what SQLAlchemy's doing under the hood.

api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,542 @@ From this example, we can observe:
...
+| `An Apple a Day` | `When Shay fell for a doctor, would a love of apples spell disaster for their happiness?` |
+<!-- prettier-ignore-end -->
+
+In response to our request, the endpoint should return success for the status code, with a list of the books that matches the filters. With a `description` value of `apple`, we'd get back the books `It's Not Easy Being an Orange` and `An Apple a Day`, but not `10,000 Apples`. If we also filter the `title` with a value of `apple`, we should get back only `An Apple a Day`. We'll still return `200 OK` for a successful response status code, and the JSON response body will be provided by the list of books.
+
+Refer to the title filtering example to review the response body structure.
+
+### Logic
+
+It's tempting to follow a pattern similar to the title filtering, but as the number of filters we want to check increases, our code would become more and more complex. Right now, or logic is structured like this:
```
kelsey-steven-ada, line 380:
```suggestion
It's tempting to follow a pattern similar to the title filtering, but as the number of filters we want to check increases, our code would become more and more complex. Right now, our logic is structured like this:
```

api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,542 @@ From this example, we can observe:
...
+| - | - |
+| `query = db.select(Book)` | Start by creating a `Select` object on the `book` table. |
+| `if title_param:` | Only run the block code if a title filter was supplied. |
+| `query = query.where(...)` | Calls the `.where()` method on the previous `Select` object (the one returned by `db.select()`), returning a new `Select` object with the conditions in the `where` applied. Update the `query` variable to refer to this new, restricted query. If the title filter was not supplied, this code would not run, and `query` would still refer to the unrestricted `Select` object from `db.select(Book)`. |
+| `query = query.order_by(Book.id)` | Whichever query the `query` variable refers to (the unrestricted `Book` query, or the updated query that includes the title filter), update it to order the results by the book IDs. |
+<!-- prettier-ignore-end -->
+
+Basically, we pealed off the shared part at the start of both queries, added a condition to filter by title if the `title` query param is present, and then ordered the results by the book's ID. This code is equivalent to the previous code, but it's structured in a way that will make it easier to add more filters in the future.
```
kelsey-steven-ada, line 436:
```suggestion
Basically, we peeled off the shared part at the start of both queries, added a condition to filter by title if the `title` query param is present, and then ordered the results by the book's ID. This code is equivalent to the previous code, but it's structured in a way that will make it easier to add more filters in the future.
```

api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,542 @@ From this example, we can observe:
...
+        query = query.where(Book.title.ilike(f"%{title_param}%"))
+
+    query = query.order_by(Book.id)
+```
+
+<!-- prettier-ignore-start -->
+| <div style="min-width:250px;"> Piece of Code </div> | Notes |
+| - | - |
+| `query = db.select(Book)` | Start by creating a `Select` object on the `book` table. |
+| `if title_param:` | Only run the block code if a title filter was supplied. |
```
kelsey-steven-ada, line 431:
```suggestion
| `if title_param:` | Only run the block of code if a title filter was supplied. |
```

api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,542 @@ From this example, we can observe:
...
+
+##### !end-answer
+
+##### !explanation
+
+- The name of the resource collection is `books`, so the path of this URL should start with `/books` rather than listing the resource type as a query param.
+- This URL makes mistakes in the separation characters. `&` is used to separate multiple query params in the query string. `?` separates the path from the query string. `=` separates the key from the value.
+- Although we are filtering by title, the resource collection should be `books`, not `titles`.
+- This example follows RESTful conventions for filtering books by title. The path is `/books`, and the query string contains a `title` query param.
+- The query param in this URL is backwards, reversing the key and the value.
```
kelsey-steven-ada, line 574: I'd suggest using numbers rather than bullets since these correspond to an exact order.

```suggestion
1. The name of the resource collection is `books`, so the path of this URL should start with `/books` rather than listing the resource type as a query param.
2. This URL makes mistakes in the separation characters. `&` is used to separate multiple query params in the query string. `?` separates the path from the query string. `=` separates the key from the value.
3. Although we are filtering by title, the resource collection should be `books`, not `titles`.
4. This example follows RESTful conventions for filtering books by title. The path is `/books`, and the query string contains a `title` query param.
5. The query param in this URL is backwards, reversing the key and the value.
```

api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,542 @@ From this example, we can observe:
...
-* id: 14ef34fa
+
+* type: multiple-choice
+* id: 7cab0a7c-d54a-4a1f-885d-6d77bc3294fe
 * title: Query Params
+
 ##### !question
 
-Think about Query Params and Filtering Books by Title.
+Which of the following URLs follows RESTful conventions for filtering books by title?
```
kelsey-steven-ada, line 548: I'm wondering if it's worth including that the example title we're filtering by is "analytical engine"

api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,542 @@ From this example, we can observe:
...
+
+- The name of the resource collection is `books`, so the path of this URL should start with `/books` rather than listing the resource type as a query param.
+- This URL makes mistakes in the separation characters. `&` is used to separate multiple query params in the query string. `?` separates the path from the query string. `=` separates the key from the value.
+- Although we are filtering by title, the resource collection should be `books`, not `titles`.
+- This example follows RESTful conventions for filtering books by title. The path is `/books`, and the query string contains a `title` query param.
+- The query param in this URL is backwards, reversing the key and the value.
+
+<br />
+
+All these examples use the text `analytical%20engine` as the filter. Recall that `%20` is the URL encoded value for a space character, so we are looking for a title containing the words `analytical engine`. It was in her notes about the Analytical Engine that Ada Lovelace described how this machine could function as a programmable computer, even describing what is widely considered to be the first computer program.
```
kelsey-steven-ada, line 578: If we don't include the title being searched in the prompt, I had the thought that this content could be useful as a hint rather than part of the explanation.

kelsey-steven-ada (commented):
I love the restructuring of headings, supporting info you've added, and continued reminders about how what we learned in SQL is applicable here. Just a few suggestions/questions overall.

I want to get student feedback before adding more branch work to our plate, but it crossed my mind that this lesson is a bit long now and we may want to consider breaking it into 2 parts in the future.

api-5-more-queries/query-params.md
anselrognlie: Per your later suggestion, I'll move the search string part of the explanation to the hint.

api-5-more-queries/query-params.md
anselrognlie: I had initially thought the same, but since the responses aren't themselves numbered, I was concerned it might be confusing.

api-5-more-queries/query-params.md
anselrognlie: I think I prefer moving this to the hint rather than explicitly stating the search string in the prompt. I like the idea of the reader needing to make sense of what they're seeing, but if they get hung up on wondering what the analytical engine stuff is, it's covered in the hint and they can move on.

api-5-more-queries/query-params.md
```diff
@@ -80,198 +103,546 @@ From this example, we can observe:
...
+- If we call `query.where()` without updating `query`, what would `query` refer to when we try to use it later?
+
+##### !end-hint
+
+##### !explanation
+
+- By updating `query` here, if we later add more conditions to the query, the later use will include the effects of the earlier conditions. If we didn't update `query`, the later use would ignore the earlier conditions.
+- It's true that `where()` returns a new `Select` object, so we would either need to use that value immediately (such as by calling another query method on it, or passing it directly to `db.session.scalars()`), or store it in a variable for later use.
+- There is no syntax error here. Python allows us to ignore return values. However, this would probably result in a logical error!
+- Failing to update `query` means that `query` would refer to the original `Select` object, which would not have the conditions applied by the `where()` method. This would result in `books` always containing all books, not just the filtered ones.
```
kelsey-steven-ada, line 643: I forgot to leave a note here initially, for the same reasons and to be consistent with the prior question, I'd like to flip these over to numbered.

```suggestion
1. By updating `query` here, if we later add more conditions to the query, the later use will include the effects of the earlier conditions. If we didn't update `query`, the later use would ignore the earlier conditions.
2. It's true that `where()` returns a new `Select` object, so we would either need to use that value immediately (such as by calling another query method on it, or passing it directly to `db.session.scalars()`), or store it in a variable for later use.
3. There is no syntax error here. Python allows us to ignore return values. However, this would probably result in a logical error!
4. Failing to update `query` means that `query` would refer to the original `Select` object, which would not have the conditions applied by the `where()` method. This would result in `books` always containing all books, not just the filtered ones.
```

kelsey-steven-ada (commented):
I forgot to call out the same numbering change for the last CFU question, everything else is looking good.

kelsey-steven-ada (approved):

PR #156, opened 2024-04-30
Building an API pt. 4 - lesson 1: "Read One Book" updates

kelsey-steven-ada:
[Pt. 4 Asana ticket](https://app.asana.com/0/1205655776549324/1206697443678569/f)
[Pt. 4 Lesson 1 ticket](https://app.asana.com/0/1205655776549324/1207017851991892/f)


api-4-read-update-delete/read-one-book-2.md
```diff
@@ -12,7 +13,7 @@ We will build our Hello Books API to fulfill the following feature:
 
 1. As a client, I want to send a request to get one existing book, so that I can see the `id`, `title`, and `description` of the book.
 
-We will refactor the endpoint we designed in our previous lesson [02) Building and API - Read One Book Endpoint](../api-2-read-404s/read-one-book.md) to make use of the data in our postgres database.
+We will refactor the `get_one_book` endpoint we designed in a prior lesson to make use of the data in our postgres database.
```
kelsey-steven-ada, line 16: We've been trying to minimize direct links between lesson that makes them more fragile, what do folks think of removing the links with this language?

api-4-read-update-delete/read-one-book-2.md
```diff
@@ -34,56 +35,82 @@ We will refactor the endpoint we designed in our previous lesson [02) Building a
...
-1. Format the book data into the appropriate structure (a single dictionary with `id`, `title`, and `description`)
-1. Send back a response
+2. Retrieve the book with the matching `book_id` from the database
+3. Format the book data into the appropriate structure (a single dictionary with `id`, `title`, and `description`)
+4. Send back a response
 
-Our new route needs to read data from the incoming request path. Our existing route doesn't do this, so we'll need to add an entirely new route to our existing `Blueprint`.
+Let's take a look at how our updated route will account for this!
 
-Our new route needs a **route parameter**. The route `/books/1` should give us the details for the book with `id` 1. `/books/2` should give us details for book with `id` 2, `/books/3000` should give us details for book 3000, and so on.
```
kelsey-steven-ada, line 66: This lesson looks like it was copy/pasted from Pt. 2 and only the necessary parts were updated. Through the lesson I've removed or re-organized fully duplicated sections like this.

api-4-read-update-delete/read-one-book-2.md
```diff
@@ -94,124 +121,182 @@ def handle_book(book_id):
...
-
-### !end-callout
+| <div style="min-width:250px;"> Piece of Code </div> | Notes |
+| --------------------------------------------------- | ----- |
+| `@books_bp.get("/<book_id>")`                       | Unchanged from our previous hardcoded route. This decorator indicates that the function below it is a `GET` route that is registered with the `Blueprint` object named `books_bp`. |
+| `"/<book_id>"`                                      | Unchanged from our previous hardcoded route. This is the `Blueprint` syntax to indicate a route parameter named `<book_id>`. The `<book_id>` placeholder shows that we are looking for a variable value which we can use in the function as the variable `book_id`. |
+| `def get_one_book(book_id):`                        | Unchanged from our previous hardcoded route. This function must have a parameter, `book_id`. This parameter name must match the route parameter in the decorator. It will receive the part of the request path that lines up with the placeholder in the route. |
```
kelsey-steven-ada, line 128: Looking for input on if these are useful or should be cut as repetitive.

api-4-read-update-delete/read-one-book-2.md
```diff
@@ -94,124 +121,182 @@ def handle_book(book_id):
...
 
+## Manually Testing the Get One Book Endpoint
+
+As with our previous routes, we should practice using all the tools at our disposal to test our endpoints and investigate issues should they arise.
+
+- Postman
+- Flask Server logs
+- `psql` interface
+- VS Code
+- Peers, classmates, and rubber ducks
```
kelsey-steven-ada, line 219: I know this section is small, but I wanted to have some shout out that we should be testing our updated endpoint. Does this seem okay, or is there another way folks might want to include that info/sentiment?

assets/building-an-api/create-and-read_get-books-postman.png
kelsey-steven-ada, line 1: When making the changes for pt. 3, I didn't realize there was an image shared between pt. 1 & pt. 3 that was moved, breaking a Pt. 1 lesson. Adding a dark mode screenshot with the same name and location to fix the broken image in pt. 1 "Read All Books"

api-4-read-update-delete/read-one-book-2.md
anselrognlie: I'm OK with removing it, but still do want to point out that Learn does validate links to other locatins in the curriculum (at least within the same unit, not sure about different units). For instance, if I add the link back in, and then rename the page it links to, Learn detects that the link is broken and displays a warning in the browser (even when I pare down the config.yaml during testing).

So while it's technically brittle, as we need to be sure to update links if we rename/move pages, Learn does help us detect the situation.

api-4-read-update-delete/read-one-book-2.md
anselrognlie: This is only the second time we're looking at the get one book route, correct? (first hard coded, and now data-driven) I'm fine with folks having another opportunity to have the route param and such reviewed. You made some text changes that acknowledges they're not reading this in a vacuum, so my inclination would be to retain it.

api-4-read-update-delete/read-one-book-2.md
anselrognlie: It's easy to get stuck and overlook something that could be helpful. Would your thought then be to continue calling these out in subsequent lessons? I could see this being a helpful reminder at first, but eventually folks will probably start skipping past them, or at the very least read them less carefully. This could result in missed materials, if there's new information presented going foward.

assets/building-an-api/create-and-read_get-books-postman.png
anselrognlie: Good call. We should probably avoid linking resources across lessons (though good that Learn noticed!) But if we're not careful, we could end up in situations where changing the contents of one image effectively changes the contents of another, but doesn't result in a break. This could be confusing. So even if two images end up just being copies of one another, making the copies should help insulate us a bit.

api-4-read-update-delete/read-one-book-2.md
kelsey-steven-ada: You make a good point about it being repetitive and skipped over eventually. I think after this lesson we can leave it off until something changes and maybe just include a single sentence after the code is done along the lines of "Make sure to manually test your endpoints to ensure everything works as expected."?

api-4-read-update-delete/read-one-book-2.md
```diff
@@ -94,124 +121,182 @@ def handle_book(book_id):
...
+##### !answer
+
+a|
+b|
+c|
+
+##### !end-answer
+##### !explanation
+
+Each of these is a valid reason for separating out our code into helper functions! This change helps our functions to better align with the 'single responsibility principle' and helps to reduce repeated code across a project, making a project easier to maintain over time.
```
anselrognlie, line 298: uses capitalization rather than scare quotes 😱. Consider bold or italics if further setting off is desired.
```suggestion
Each of these is a valid reason for separating out our code into helper functions! This change helps our functions to better align with the Single Responsibility Principle and helps to reduce repeated code across a project, making a project easier to maintain over time.
```

api-4-read-update-delete/read-one-book-2.md
```diff
@@ -94,124 +121,182 @@ def handle_book(book_id):
...
-* Planned the HTTP response, request, and logic for this endpoint
-* Considered route parameters, and how we need the book ID in the route
-* Created a new endpoint that matches on `GET` requests to `"/<book_id>"`
-* Defined this endpoint with the function signature `def book(book_id):`
-* Used `Book.query.get(book_id)` to get an instance of `Book` matching `book_id`
-* Returned a dictionary literal as our response
-* Handled `400` and `404` responses
+a| Having a function with the single responsibility of checking if an `id` is valid then retrieving the `Book` record.
+b| Moving data validation and fetching out of the function `get_one_book`, so that `get_one_book` is focused on the unique work for the route.
+c| Knowing we will create further routes which will need to look up a `Book` record by `id`.
```
anselrognlie, line 286: I'm not sure why, but something about the way the prompt is worded really made the gerund-phrased answers sound really strange to me. I tried rewording the answers a bit.
```suggestion
a| Separating out the single responsibility of retrieving valid `Book` records allows us to encapsulate the validation logic, which could be complex, in a function with a self-documenting name.
b| Moving data validation and fetching out of the function `get_one_book`allows `get_one_book` to focus on the unique work for the route.
c| We know that we will create further routes which will need to look up a `Book` record by `id`. 
```

api-4-read-update-delete/read-one-book-2.md
anselrognlie: Sounds good to me.

anselrognlie (changes requested):
Looking good. A few rephrasing suggestions, and there may be a few opportunities to flesh out explanations with additional paragraphs/callouts.

api-4-read-update-delete/read-one-book-2.md
```diff
@@ -94,124 +121,182 @@ def handle_book(book_id):
...
 </details>
 
+## Manually Testing the Get One Book Endpoint
+
+As with our previous routes, we should practice using all the tools at our disposal to test our endpoints and investigate issues should they arise.
+
+- Postman
+- Flask Server logs
+- `psql` interface
+- VS Code
```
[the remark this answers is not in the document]
kelsey-steven-ada: VS Code was in prior lists of tools to use when testing through the lesson series, my thinking was it was likely about placing print statements in routes so you can see info in the console.

api-4-read-update-delete/read-one-book-2.md
```diff
@@ -94,124 +121,182 @@ def handle_book(book_id):
...
+| `def get_one_book(book_id):`                        | Unchanged from our previous hardcoded route. This function must have a parameter, `book_id`. This parameter name must match the route parameter in the decorator. It will receive the part of the request path that lines up with the placeholder in the route. |
+| `query = ...`                                       | We store the `Select` object which represents our query for a single `Book` instance in the variable `query` |
+| `... = db.select(Book) ...`                         | This syntax tells the `SQLAlchemy` object `db` to select all entries for the model `Book`. This method returns a `Select` object that represents the query we are building. |
+| `....where(Book.id == book_id)`                     | This is an example of building up a query by adding a where clause onto a `Select` object. We call the `where` method and pass it a comparison `Book.id == book_id` to use for filtering the results of our initial `Select` statement created by `db.select(Book)`. The `where` function will return a new `Select` object that is stored in the variable `query` |
```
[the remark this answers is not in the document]
anselrognlie: I always get nervous with links to SqlAlchemy docs. They look different enough from how we access SqlAlchemy through the flask extension that it could be confusing. There's a lot of mental rearrangement we have to do that is easy to take for granted. But I do like their note that
> For boolean expressions, most Python operators such as ==, !=, <, >= etc. generate new SQL Expression objects, rather than plain boolean True/False values

So your plan sounds good to me.

api-4-read-update-delete/read-one-book-2.md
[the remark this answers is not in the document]
anselrognlie: Do you know where the first time it was mentioned whether there was any additional context to that effect? If that purpose was established once elsewhere, then it feels fine to not mention it again here. But if we never said what using VSCode as a debug tool means, it would be nice to have that added somewhere.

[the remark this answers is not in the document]
kelsey-steven-ada: It shows up first in Pt. 1 "Read All Books", but we don't have additional context, it's just a bulleted list in an info callout box. I'll add a ticket to the backlog to add more context.

api-4-read-update-delete/read-one-book-2.md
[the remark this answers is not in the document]
kelsey-steven-ada: I added the following text higher in the lesson, right where we show the `query` declaration. It felt like a good space to tie it directly to the syntax as we introduced it without being squished in a table or slightly separated from code by a callout down here, but I'd love to hear what you think:

> Let's take a deeper look at our `where` method's parameter `Book.id == book_id`. In plain SQL we might have expressions in a WHERE clause such as `id = 3` or `title = Fictional Book`. SQLAlchemy allows us to compose SQL expressions like this by using the standard Python operators in conjunction with SQLAlchemy's `Column` class. For boolean expressions, most Python operators such as ==, !=, <, >= etc. generate new SQL Expression objects, rather than plain boolean True/False values. 
> 
> Re-reading our expression `Book.id == book_id`, we now know that this statement does not evaluate to a True/False value. `Book.id == book_id` is generating a SQL Expression object that can be applied when we execute `query`, where `Book.id` is a SQLAlchemy `Column` object representing the `id` column of our `Book` model.

api-4-read-update-delete/read-one-book-2.md
```diff
@@ -94,124 +125,184 @@ def handle_book(book_id):
...
-def read_one_book(book_id):
-    book = validate_book(book_id)
-    return {
-            "id": book.id,
-            "title": book.title,
-            "description": book.description
-        }
 ```
 
+Even though `scalar` can return `None`, we don't need to worry about that case in our route function because it is already handled! `validate_book` takes care of that scenario by sending back a 404 response with `abort`, which prevents the rest of the route code from running.
```
anselrognlie, line 213: 👍

anselrognlie (approved):
LGTM. Just added some minor suggestions to the new blurb about the `where` behavior.

PR #160, opened 2024-05-29
Fix up Hello World Routes Lesson

yangashley:
Remove references to routes.py, add a section to create hello_world_routes.py, fix up typos

### DIFF
--- a/api-1-setup-read/hello-world-routes.md
+++ b/api-1-setup-read/hello-world-routes.md
@@ -23,18 +23,21 @@ We will:
 
 ## Defining Endpoints with Blueprint
 
-Our code will import and use something named `Blueprint` from `flask`. In `routes.py`, let's add this import line.
-
-```python
-from flask import Blueprint
-```
-
 `Blueprint` is a Flask class that provides a pattern for grouping related routes (endpoints). Flask will often refer to these routes using the word "view" due to Flask having the potential of sending HTML views. However, we will be sending back JSON.
 
 For now, let's focus on how we can use a `Blueprint` in our own code, but for future reference, we should consider:
 
 - [Flask's definition of `Blueprint`](https://flask.palletsprojects.com/en/3.0.x/api/#blueprint-objects)
 - [Flask's tutorial on using `Blueprint`](https://flask.palletsprojects.com/en/3.0.x/tutorial/views/)
+
+### Creating a Our First Route File 
+
+In the `routes` directory, we will create our first route file in the Hello Books API project. We can use the following command from the root of our project directory:
+
+```
+$ touch app/routes/hello_world_routes.py
+```
+
 
 ### Creating a Blueprint in `hello_world_routes.py`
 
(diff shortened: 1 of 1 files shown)


api-1-setup-read/hello-world-routes.md
```diff
@@ -133,8 +136,7 @@ hello_world_bp = Blueprint("hello_world", __name__)
 
 @hello_world_bp.get("/")
 def say_hello_world():
-    my_beautiful_response_body = "Hello, World!"
-    return my_beautiful_response_body
+    return "Hello, World!"
```
yangashley, line 139: Changed the body of this route to match what is shown in branch 01c-hello-world: https://github.com/AdaGold/hello-books-api/blob/01c-hello-world/app/routes/hello_world_routes.py#L6-L8

api-1-setup-read/hello-world-routes.md
```diff
@@ -23,19 +23,22 @@ We will:
 
 ## Defining Endpoints with Blueprint
 
-Our code will import and use something named `Blueprint` from `flask`. In `routes.py`, let's add this import line.
-
-```python
-from flask import Blueprint
-```
-
```
kelsey-steven-ada, line 31: It feels like we're skipping back and forth a little bit by introducing blueprints, pausing to create the file, then working with blueprints. How would you feel about moving up the "Creating Our First Route File" section to the top here, then the blueprint topics can flow from there? 

If there's some dissonance between the overall title and that first section, we could add a sentence like "Before we can define any endpoints, we need a file for our route functions."

kelsey-steven-ada (commented):
Thank you for jumping on these changes while they're fresh in mind! Left a suggestion around content order.

api-1-setup-read/hello-world-routes.md
kelsey-steven-ada: Looks great!

api-1-setup-read/hello-world-routes.md
```diff
@@ -21,14 +21,18 @@ We will:
...
 
-```python
-from flask import Blueprint
+Before we can define any endpoints, we need a file for our route functions. In the `routes` directory, we will create our first route file in the Hello Books API project. We can use the following command from the root of our project directory:
+
+```
+$ touch app/routes/hello_world_routes.py
 ```
 
+### Defining Endpoints with Blueprint
```
kelsey-steven-ada, line 34: I'm sorry, I didn't catch the headers when looking at the content placement in the screenshot. The content under this heading is info about Blueprint but doesn't cover defining endpoints, which feels potentially confusing when trying to find info on the page. I think if we swap the headers "Blueprints in Flask" & "Defining Endpoints with Blueprint" they would better match what follows, what do you think?

yangashley: no prob! I swapped the order like you suggested so "Defining Endpoints with Blueprint" comes first, then "Creating Our First Route File", then "Blueprints in Flask", then "Creating a Blueprint in `hello_world_routes.py`.

I think that does flow a little better, thanks!

kelsey-steven-ada (approved):

PR #161, opened 2024-06-03
updates for api pt 4: updates

mikellewade:
Updates for Building an API Pt. 4: Updates 

[Building API Pt. 4 Asana Ticket ](https://app.asana.com/0/1201700438643002/1207017851991894)
[Building API Pt. 4: Updates Asana Ticket](https://app.asana.com/0/1203422600663187/1207017851991894)


api-4-read-update-delete/update.md
```diff
@@ -1,6 +1,6 @@
 # Update
 
-<iframe src="https://adaacademy.hosted.panopto.com/Panopto/Pages/Embed.aspx?pid=9594d398-a406-4a5a-9a9a-ae6901622f0f&autoplay=false&offerviewer=true&showtitle=true&showbrand=true&captions=true&interactivity=all" height="405" width="720" style="border: 1px solid #464646;" allowfullscreen allow="autoplay"></iframe>
+<!-- <iframe src="https://adaacademy.hosted.panopto.com/Panopto/Pages/Embed.aspx?pid=9594d398-a406-4a5a-9a9a-ae6901622f0f&autoplay=false&offerviewer=true&showtitle=true&showbrand=true&captions=true&interactivity=all" height="405" width="720" style="border: 1px solid #464646;" allowfullscreen allow="autoplay"></iframe> -->
```
kelsey-steven-ada, line 3: Just missing the marker we're adding to find these later when we're able to replace the videos.

```suggestion
<!-- FLASK UPDATE -->
<!-- <iframe src="https://adaacademy.hosted.panopto.com/Panopto/Pages/Embed.aspx?pid=9594d398-a406-4a5a-9a9a-ae6901622f0f&autoplay=false&offerviewer=true&showtitle=true&showbrand=true&captions=true&interactivity=all" height="405" width="720" style="border: 1px solid #464646;" allowfullscreen allow="autoplay"></iframe> -->
```

api-4-read-update-delete/update.md
```diff
@@ -42,52 +42,50 @@ This assumes that there is a `book` table with at least the following row:
 | ---- | ---------------------- | ------------------------------------------- |
 | `1`  | `Fictional Book Title` | `A fantasy novel set in an imaginary world` |
 
-When the record is successfully updated, we should return the appropriate status code, which is `200 OK`. As with other endpoints, we may choose our response body from a variety of possibilities, but for simplicity, we will return a brief status message.
+When the record is successfully updated, we should return the appropriate status code, which is `204 NO CONTENT`. The `204` is a successful status response code that is used when a request has succeeded and there is no response body. The full response for this route will be `204` status code accompanied by an empty response body.
 
 | Response Status | Response Body                  |
 | --------------- | ------------------------------ |
-| `200 OK`        | `Book #1 successfully updated` |
+| `204 NO CONTENT`        ||
```
kelsey-steven-ada, line 49: For the delete lesson we added a comment like the one below for the response body side, we could do something similar here. (The added div content is to make sure the first "Response Status" column doesn't get collapsed and stack the words vertically)

```suggestion
| <div style="min-width:200px;"> Response Status </div> | Our successful response to a `PUT` request will not have a response body |
```

api-4-read-update-delete/update.md
```diff
@@ -42,52 +42,50 @@ This assumes that there is a `book` table with at least the following row:
...
-1. Update the instance of `Book` with the new data
-1. Save the updated `Book` in the database
-1. Send back a response
+2. Retrieve the `Book` instance with the matching `book_id` from the database
+3. Read the new, updated book data from the HTTP request
+4. Update the instance of `Book` with the new data
+5. Save the updated `Book` in the database
+6. Send back a response
 
 ## Updating a Book Endpoint: Code
```
kelsey-steven-ada, line 62: I think it might be worth adding a short section above the code to introduce that we are creating a response in a different way than students have seen before, perhaps as with it's own header, depending on size. 

Top of mind, I think it would be useful to point out:
1.  unlike returning a tuple or using make_response, we're directly instantiating a `Response` object because we need the level of control over the response that offers us. 
2. If we didn't use Response and set the mimetype, Flask would default to an HTML response and would introduce different response types between the endpoints of our API

api-4-read-update-delete/update.md
```diff
@@ -42,52 +42,50 @@ This assumes that there is a `book` table with at least the following row:
...
 | `book.title = form_data["title"]`                   | We'll use our OOP skills to update `book`'s `title` attribute                                                                              |
 | `book.description = ...`                            | We'll use our OOP skills to update `book`'s `description` attribute                                                                        |
 | `db.session.commit()`                               | Every time a SQLAlchemy model has been updated, and we want to commit the change to the database, we'll execute `db.session.commit()`.     |
-| `return make_response( ... )`                            | This is one of many ways we can return our appropriate HTTP response. Since we didn't supply a status code, Flask will default to `200 OK`.                                                                      |
+| `return Response(status=204, mimetype="application/json")`                            | This is one of many ways we can return our appropriate HTTP response. We use the `Response` class to create a response to send back with a status code of `204` and a data type of `application/json`.
```
kelsey-steven-ada, line 88: With linebreaks in the code and description, this might be a little easier to consume if we break it down further. I used the following breakdown for the delete endpoint in case you'd like to do something similar:

```
| `return Response(...)` | We can use the `Response` constructor to manually create a `Response` object when we need detailed control over the contents and attributes of the endpoint's response. |
| `(status=204, ...)` | As we did in the `update_book` endpoint, when calling the `Response` constructor, we use the keyword argument `status` to pass a status code of `204` for our endpoint's response. |
| `(..., mimetype="application/json")` | As we did in the `update_book` endpoint, we need to set the `mimetype` for our response to `"application/json"` so that the response type does not default to `HTML`. We do this to ensure our API has a consistent response type across our endpoints. |
```

api-4-read-update-delete/update.md
```diff
@@ -138,5 +136,40 @@ Check off all the topics that we've briefly touched on so far.
...
+
+### !challenge
+
+* type: multiple-choice
+* id: be152c53-2f78-40e1-8c6c-9b3f8q086c17
+* title: Hello World Routes
+
+##### !question
+
+Which status code should be used if response is returned with an empty response body?
```
kelsey-steven-ada, line 149:
```suggestion
Which status code should be used if a response is returned with an empty response body?
```

kelsey-steven-ada (commented):
Thank you for taking this piece on! 😊 I left a couple suggestions and some thoughts on how we could dive a bit further into the Response object.

api-4-read-update-delete/update.md
mikellewade: Oh thank you so much! Just implemented this suggestion!

api-4-read-update-delete/update.md
```diff
@@ -42,53 +43,57 @@ This assumes that there is a `book` table with at least the following row:
 | ---- | ---------------------- | ------------------------------------------- |
 | `1`  | `Fictional Book Title` | `A fantasy novel set in an imaginary world` |
 
-When the record is successfully updated, we should return the appropriate status code, which is `200 OK`. As with other endpoints, we may choose our response body from a variety of possibilities, but for simplicity, we will return a brief status message.
+When the record is successfully updated, we should return the appropriate status code, which is `204 NO CONTENT`. The `204` is a successful status response code that is used when a request has succeeded and there is no response body. The full response for this route will be `204` status code accompanied by an empty response body.
 
 | Response Status | Response Body                  |
 | --------------- | ------------------------------ |
-| `200 OK`        | `Book #1 successfully updated` |
+| <div style="min-width:200px;"> Response Status </div> | Our successful response to a `PUT` request will not have a response body |
```
kelsey-steven-ada, line 50: It looks like we lost the `204` in the change but GitHub isn't allowing me to add it as a suggestion on the lines, below is what I think we wanted here.

```
| <div style="min-width:200px;"> Response Status </div> | Response Body |
| --------------- | ------------------------------ |
| `204 NO CONTENT` | Our successful response to a `PUT` request will not have a response body |
```

api-4-read-update-delete/update.md
```diff
@@ -42,53 +43,57 @@ This assumes that there is a `book` table with at least the following row:
...
+4. Update the instance of `Book` with the new data
+5. Save the updated `Book` in the database
+6. Send back a created response
 
 ## Updating a Book Endpoint: Code
 
-This endpoint uses the same path as our existing route for reading a `Book` record, `"/<book_id>"`. We could refactor and expand on this same function. However, just as we did with `create` and `read` for the `/books` route, we'll create a seperate route function fo the update functionality. 
+This endpoint uses the same path as our existing route for reading a `Book` record, `"/<book_id>"`. We could refactor and expand on this same function. However, just as we did with `create` and `read` for the `/books` route, we'll create a seperate route function fo the update functionality. In this route we will also need to create a response.
+
+## Creating a Response Object
```
kelsey-steven-ada, line 67: This feels like a subsection of `Updating a Book Endpoint: Code`, so I think we could use an h3 header here.

```suggestion
### Creating a Response Object
```

api-4-read-update-delete/update.md
```diff
@@ -42,53 +43,57 @@ This assumes that there is a `book` table with at least the following row:
...
 | `db.session.commit()`                               | Every time a SQLAlchemy model has been updated, and we want to commit the change to the database, we'll execute `db.session.commit()`.     |
-| `return make_response( ... )`                            | This is one of many ways we can return our appropriate HTTP response. Since we didn't supply a status code, Flask will default to `200 OK`.                                                                      |
-
+| `return Response(...)` | By using the `Response` constructor, we can manually create a `Response` object when we need detailed control over the contents and attributes of the endpoint's response. |
+| `(status=204, ...)` | The `Response` constructor allows us to use the keyword argument `status` to set a status code of `204` for our endpoint's response. |
+| `(..., mimetype="application/json")` | Since we construct our own response object we need to set the `mimetype` for our response to `"application/json"`. This is so that the response type does not default to `HTML` and ensure our API has a consistent response type across our endpoints. |
```
kelsey-steven-ada, line 96:
```suggestion
| `(..., mimetype="application/json")` | Since we construct our own response object we need to set the `mimetype` for our response to `"application/json"`. This is so that the response type does not default to `HTML` and ensures our API has a consistent response type across our endpoints. |
```

kelsey-steven-ada (approved):
LGTM!

PR #162, opened 2024-06-07
Building an API pt. 6 - lesson 4: Test Setup

kelsey-steven-ada:
[Pt. 6 Lesson 4 ticket](https://app.asana.com/0/1205655776549324/1207017851991899/f)


api-6-testing/test-setup.md
```diff
@@ -53,18 +54,37 @@ Let's take a look at what these files will do:
...
 ## `conftest.py`
 
+### Managing Dependencies
+
+Our `conftest.py` file will be responsible for setting up the test configuration for our app. To ensure that our environment variables are loaded from our `.env` file at the appropriate time, we need the help of a package called `python-dotenv`. The `python-dotenv` package (often referred to as simply "dotenv") is a popular, well-used tool for managing environment variables. 
+
+We should ensure that `python-dotenv` is installed by either:
+
+1. Checking our `requirements.txt` and confirming `python-dotenv` is already listed, or
+2. Installing it with `(venv) $ pip install python-dotenv`
```
yangashley, line 64: I know that the bash command you have listed shows `(venv)` to indicate that your virtual environment is activated. I'm not sure how evident that is to students. What do you think about adding a note to the instruction? Something like this:

2. Installing it with `(venv) $ pip install python-dotenv`. Notice that `(venv)` indicates that the virtual environment for our project is activated

api-6-testing/test-setup.md
```diff
@@ -83,23 +103,30 @@ def client(app):
...
 |`from flask.signals import request_finished`|We use the `@request_finished` decorator to create a new database session after a request as described below. | 
-| `@pytest.fixture`                                   | We'll create and use a pytest fixture named `app`, which will be used in our `client` fixture (defined later)                                                                                              |
-| `app = create_app( ... )`                           | When we run our tests, this line will run and create an `app` object. It's using the same `create_app` function defined in our `app/__init__.py` file!                                                      |
-| `{"TESTING": True}`                                 | Here, we're passing in a dictionary to represent a "test config" object. The current implementation of `create_app()` in `app/__init__.py` uses this argument only to check if it's truthy (which this is!). |
+| `from dotenv import load_dotenv` | We'll use `load_dotenv` to manually load the contents of our `.env` into our environment variables. |
+| `import os` | As we did in the root __init__.py we'll use os to read our environment variables. |
```
yangashley, line 113:
```suggestion
| `import os` | As we did in the root `__init__.py` we'll use `os` to read our environment variables. |
```

api-6-testing/test-setup.md
```diff
@@ -83,23 +103,30 @@ def client(app):
...
 |`db.session.remove()`| After a request is made in our test, this line creates a new database session so that we can test that changes were persisted in the database. This is particularly relevant for testing the update method. Without this line, when we test that an update was made to a record following a `put` request, the test will only look at the in-memory copy of the updated record. By adding `db.session.remove()`, we make sure the test checks that the update was persisted in the database.
-| `with app.app_context():`                           | This syntax designates that the following code should have an _application context_. This lets various functionality in Flask determine what the current running app is. This is particularly important when accessing the database associated with the app.                                                                   |
+| `with app.app_context():` | This syntax designates that the following code should have an _application context_. This lets various functionality in Flask determine what the current running app is. This is particularly important when accessing the database associated with the app. |
```
yangashley, line 123: Don't feel super strongly about this point because we're already introducing many new concepts in this lesson, but we could link "application context" for folks that want to read more on it: https://flask.palletsprojects.com/en/3.0.x/appcontext/

api-6-testing/test-setup.md
kelsey-steven-ada: Hmm...I'd be curious to hear other instructor's opinions, we have callouts in earlier lessons about the command line prompt, by this point students have completed several projects and half of the Flask lesson which require students to use a virtual environment, so I'm not sure how useful a call out would be here.

api-6-testing/test-setup.md
kelsey-steven-ada:
```suggestion
| `import os` | As we did in the root `__init__.py` file, we'll use `os` to read our environment variables. |
```

api-6-testing/test-setup.md
kelsey-steven-ada: I like linking the doc in case folks want to dig in more.

api-6-testing/test-setup.md
yangashley: Ok, that works for me if you want to leave it as is. My hope is that students see `(venv)` and remember what it means

yangashley (approved):

PR #164, opened 2024-06-13
Building an API pt. 6 - lesson 6: GET /book/<book_id> Test

kelsey-steven-ada:
[Pt. 6 Lesson 6 ticket](https://app.asana.com/0/1205655776549324/1207017851991903/f)


api-6-testing/read-one-book-test.md
```diff
@@ -226,25 +249,45 @@ Success! We passed this test. Congratulations! 🎉
...
+* type: checkbox
 * id: 72e74eca-43e2-47f2-b013-41eb3a7cbac2
-* title: Get One Book Test
+* title: GET /book/<book_id> Test
 ##### !question
 
-Think about the test for `GET` `/books/<book_id>`
-
-Check off all the topics that we've touched on so far.
+What benefits do we get from writing fixtures in `conftest.py`? Select all of the options below which apply.
```
anselrognlie, line 261: The phrasing here makes me think that it's only asking about the relevance of placing the fixtures in `conftest` (which would just cover the ability to use the fixture across many different tests). Could we reword this a little?

Alternatively, what if we remove the conftest mention from here, then add a true/false question about how putting our fixtures in conftest allows them to be used from any test file. The explanation could reiterate that fixtures aren't required to be put there, but since it's read automatically when pytest runs our tests, any test file has access to the fixtures declared there.

I suppose we could even leave the "used by many tests" here, since even a fixture defined in a test file could be used by more than one test in that file? Maybe all we need to do is drop the conftest reference from this challenge, as long as we follow upwith the second challenge?

api-6-testing/read-one-book-test.md
```diff
@@ -226,25 +249,45 @@ Success! We passed this test. Congratulations! 🎉
...
+##### !hint
+
+What actions have we seen taken with fixtures in the lesson?
+How do tests use a fixture?
+
+##### !end-hint
+##### !explanation
+
+Fixtures are great for reusable data set up. They can store data in the test database before a test runs and, if we choose, they can handle "tear down" by removing the test data in the test database to ensure each test run starts from a clean, expected state. 
+
```
anselrognlie, line 288:
```suggestion

<br />

```

anselrognlie (changes requested):
A few brief suggestions, and the new images look a little funny on my monitor. Added suggestions to deal with that.

anselrognlie (approved):
LGTM!

PR #165, opened 2024-06-14
Building an API pt. 6 - lesson 7: POST /books Test

kelsey-steven-ada:
[Pt. 6 Lesson 7 ticket](https://app.asana.com/0/1205655776549324/1207017851991904/f)


api-6-testing/post-books-test.md
anselrognlie, line 1: Super minor, but for the table row with the test declaration (around line 40), the wrapping lands so that the final `):` ends up on its own line looking very sad. Could we add a zdero width space after the open paren so that the wrap can happen there instead?

Also code quotes pytest

```md
|`def test_create_one_book(`&ZeroWidthSpace;`client):` | Continuing our best `pytest` practices, this test should start with the name `test_`, and it should describe the nature of this test.|
```

api-6-testing/post-books-test.md
```diff
@@ -37,67 +41,50 @@ def test_create_one_book(client):
 |`client.post("/books", json=...`|Sends an `POST` request to `/books` with the JSON request body passed in as the `json` keyword argument|
```
anselrognlie, line 41: Rewords to emphasize the distinction between the actual dict and the JSON it will become
```suggestion
|`client.post("/books", json=...`|Sends a `POST` request to `/books`, with the `dict` to be used as the JSON request body passed in using the `json` keyword argument| 
```

api-6-testing/post-books-test.md
```diff
@@ -37,67 +41,50 @@ def test_create_one_book(client):
 |`client.post("/books", json=...`|Sends an `POST` request to `/books` with the JSON request body passed in as the `json` keyword argument| 
 | `response_body = response.get_json()`| Get the JSON response body with `response.get_json()`|
```
anselrognlie, line 42: Rewords to emphasize the distinction between the JSON the response contained and the value we get in python
```suggestion
| `response_body = response.get_json()`| Gets the JSON response body as a Python value. Since the create logic returns a `dict`, we should anticipate `response_body` to receive a `dict` result here.|
```

api-6-testing/post-books-test.md
```diff
@@ -37,67 +41,50 @@ def test_create_one_book(client):
...
 |`assert response.status_code == 201`|Check for the expected status code|
```
anselrognlie, line 43: follows tense established in "sends a post" (combining with the next suggestion gave a very strange git diff)
```suggestion
|`assert response.status_code == 201`|Checks for the expected status code|
```

api-6-testing/post-books-test.md
```diff
@@ -37,67 +41,50 @@ def test_create_one_book(client):
...
-## Run and Debug Test
-
-When we run are test using `pytest` we will see a failure:
-
-![POST /books test failure](../assets/post_test_failure.png)
-
-Reading the test failure output indicates tat `response.get_json()` returns `None` whereas we expect it to return `"Book New Book successfully created"`.
-
-To investigate this error we may start by looking at the [`get_json` documentation](https://flask.palletsprojects.com/en/2.0.x/api/?highlight=get_json#flask.Response.get_json) and will note this piece:
+|`assert response_body == {...}}`| Check for the expected key/value pairs in the response body|
```
anselrognlie, line 44: follows tense established in "sends a post"
```suggestion
|`assert response_body == {...}}`| Checks for the expected key/value pairs in the response body|
```

api-6-testing/post-books-test.md
```diff
@@ -37,67 +41,50 @@ def test_create_one_book(client):
...
-Alternatively, we can `jsonify` the response body in our `POST` `/books` route:
-
-```python
-return make_response(jsonify(f"Book {new_book.title} successfully created"), 201)
-```
+As always, we should run `pytest` and ensure that the `POST` test passes and no other tests have been affected by our changes.
 
-Note that `POST` test passes with this change:
+![Screenshot of pytest test result: 3 tests in tests/test_book_routes.py passed](../assets/api-6-testing/api-6-testing_post_test_success.png)  
+_Fig. Test output reporting 3 tests in `tests/test_book_routes.py` passed. ([Full size image](../assets/api-6-testing/api-6-testing_post_test_success.png))_
```
anselrognlie, line 49: Thank you!

api-6-testing/post-books-test.md
```diff
@@ -37,67 +41,50 @@ def test_create_one_book(client):
...
+##### !hint
+
+What does the client fixture do for us? When do we need fixtures which create data and why?
+
+##### !end-hint
+##### !explanation
+
+1. False: The `client` fixture does not create any `Book` records
+2. False: The `client` fixture allows us to make requests to our API, so we must include it as a dependency.
+3. True: We only need a fixture which creates a `Book` record before the test runs if the test depends on there being data in the database.
```
anselrognlie, line 85: uses "that" for a restrictive clause
```suggestion
3. True: We only need a fixture that creates a `Book` record before the test runs if the test depends on there being data in the database.
```

anselrognlie (changes requested):
A few suggestions. Thanks for already including the image caption and link! 🙂

api-6-testing/post-books-test.md
```diff
@@ -28,76 +28,63 @@ def test_create_one_book(client):
...
 
-* Examined the `test_create_one_book` syntax
-* Considered using `get_data(as_text=True)`
-* Refactored our `POST` `\books`route to use `jsonify`
-* Refactored our `PUT` `\books` route to use `jsonify`
-* Refactored our `DELETE` `\books` route to use `jsonify`
+a| The `client` fixture also creates data used by the test `test_create_one_book`.
+b| `test_create_one_book` doesn't need the `client` fixture.
+c| `test_create_one_book` doesn't need to retrieve a pre-existing record, so it does not need a fixture that saves Book records.
+d| The `client` fixture needs to used for every test using the `pytest` framework.
```
kelsey-steven-ada, line 68:
```suggestion
d| The `client` fixture needs to be used for every test using the `pytest` framework.
```

api-6-testing/post-books-test.md
```diff
@@ -4,7 +4,7 @@
 
 Our goals for this lesson are to:
 - Use fixtures to create test data
```
anselrognlie, line 6: I was thinking to completely remove this one about fixtures, since this lesson doesn't use fixtures to create any test data this time. If we still want multiple goals, maybe this could be

```suggestion
- Use fixtures to access the test client to send a POST request
```

But that seems wordy to me.

kelsey-steven-ada: Ahh, I get what you're saying, I think it's fine to only have one goal and remove the fixture related one since we aren't writing new ones or doing anything new with them.

PR #170, opened 2024-08-13
Building an API pt. 8 - lesson 1: One-to-Many

kelsey-steven-ada:
[Pt. 8 Asana ticket](https://app.asana.com/0/1205655776549324/1206697443678574/f)
[Pt. 8 Lesson 1 ticket](https://app.asana.com/0/1205655776549324/1207017851991911/f)


api-8-relationships-in-sqlalchemy/one-to-many.md
```diff
@@ -198,17 +255,41 @@ class Book(db.Model):
...
 </details>
 </br>
 
-Now we have our `author_id` in the `Book` model, but what is the new `books` attribute in `Author`? And what purpose does `back_populates` serve? Refer to the [SQLAlchemy documentation](https://docs.sqlalchemy.org/en/14/orm/basic_relationships.html#one-to-many) or your favorite search engine to find an answer.
+### !callout-info
+
+## Fixing type errors in VSCode
+
+Our new code will compile and run, but sometimes our tools can get a little confused about syntax and underline something that really doesn't need to change. In this case, we may see `Pylance` in VSCode add underlines after we've created the relationship between our models. If we would like to remove those warning underlines, we can add the following code to the models:
+```py
```
anselrognlie, line 274: Thanks for including these!

api-8-relationships-in-sqlalchemy/one-to-many.md
```diff
@@ -182,14 +235,18 @@ class Author(db.Model):
...
-    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
-    title = db.Column(db.String)
-    description = db.Column(db.String)
-    author_id = db.Column(db.Integer, db.ForeignKey('author.id'))
-    author = db.relationship("Author", back_populates="books")
+    id: Mapped[int] = mapped_column(primary_key=True, autoincrement=True)
+    title: Mapped[str]
+    description: Mapped[str]
+    author_id: Mapped[Optional[int]] = mapped_column(ForeignKey("author.id"))
+    author: Mapped[Optional["Author"]] = relationship(back_populates="books")
```
anselrognlie, line 249: Would there be any value in dwelling on _why_ we made the author data optional? Like I'm cncerned students will assume that relationships should _always_ be optional rather than considering whether that makes sense for their data model.

Like here, it's primarily because we're too lazy to deal with existing Book data that has no author, and we could conceptualize NULL author data as representing a book with an unknown author (alternatively this could be represented with an "Unknown" author record). Probably don't want to open the can of worms that is books with multiple authors, as that would transform the relationship to a many-to-many, but I still feel like giving a sense of what's _required_ for a relationship as opposed to what's specific to this relationship could help.

api-8-relationships-in-sqlalchemy/one-to-many.md
```diff
@@ -230,24 +311,41 @@ If all else fails, it is okay to delete the migrations directory and the databas
...
+b|
+d|
+
+##### !end-answer
+##### !hint
+
+What is the difference between `author` and `author_id` in the `Book` class?
+What is the difference between a `ForeignKey` and a `relationship` in SQLAlchemy?
+
+##### !end-hint
```
anselrognlie, line 343: How about a second hint to help establish that the stuff realted to `relationship` aren't required?
```suggestion
##### !hint

What is the difference between `author` and `author_id` in the `Book` class?
What is the difference between a `ForeignKey` and a `relationship` in SQLAlchemy?

##### !end-hint

##### !hint

SqlAlchemy `relationship` entries provide convenient access to related objects, but aren't themselves required to establish the actual model relationship.

##### !end-hint
```

anselrognlie (changes requested):
A few small edits, and some potential areas to provide additional background.

api-8-relationships-in-sqlalchemy/one-to-many.md
kelsey-steven-ada: I think it's worth at least a brief mention, as you mentioned, it could be helpful to be clear that we don't always prefer optional relationships.

api-8-relationships-in-sqlalchemy/one-to-many.md
```diff
@@ -83,71 +102,90 @@ Refer back to [03) Building an API - Read All Books](../api-3-database-models-re
...
+
+<br/>
+
+Try out writing tests on your own, then check out the test suite we put together in GitHub on the branch for this lesson, `08a-author-model`.
+
+### !end-callout
+
 ## Visualizing and Implementing the Relationship
 
 Let's visualize the relationship between `book`s and `author`s. Just like we did when we built relationships in SQL, we will use an **entity relationship diagram** to visualize the relationship between `book`s and `author`s.
```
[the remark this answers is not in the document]
kelsey-steven-ada: Thanks for catching the properties mismatch, I can get that updated. Do you know if we've been using Figma or a different tool to create ERDs for Learn?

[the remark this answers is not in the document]
kelsey-steven-ada: Ah, looks like we were using draw.io/diagrams.net. I've made a folder in drive for us to collect curriculum assets like this that don't already live in Figma.

api-8-relationships-in-sqlalchemy/one-to-many.md
```diff
@@ -83,71 +102,90 @@ Refer back to [03) Building an API - Read All Books](../api-3-database-models-re
...
-            }
-        )
-    return jsonify(authors_response)
+@bp.get("")
+def get_all_authors():
+    query = db.select(Author)
+
+    name_param = request.args.get("name")
+    if name_param:
+        query = query.where(Author.name.ilike(f"%{name_param}%"))
```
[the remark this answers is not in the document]
kelsey-steven-ada: Adding comment to line 99 above right after stating they can check out lesson 3 for basic set up.

api-8-relationships-in-sqlalchemy/one-to-many.md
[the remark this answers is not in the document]
kelsey-steven-ada: The table names in the original image (`books` and `authors`) don't match the tables that are created locally since our classes are `book` and `author`. How do you feel about me updating the names in the ERDs to match what students would see generated?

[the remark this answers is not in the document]
anselrognlie: Back in the day, we had a best practice of using tools that could export a native version of the eventual png and including them in the Learn material to make it easier to edit assets in the event of changes (such as .fig files from figma and .xcf files from GIMP) but unfortunately that fell by the wayside.

Thanks for making sure this lives in a shared drive, as a lot of times, the web-based assets ended up in someone's personal drive, now inaccessible to us.

api-8-relationships-in-sqlalchemy/one-to-many.md
kelsey-steven-ada: Added a callout box around line 199, let me know what you think =]

api-8-relationships-in-sqlalchemy/one-to-many.md
```diff
@@ -55,123 +58,183 @@ An `author` should have the following attributes with the specified types:
...
 
-How do we define this foreign key in our Flask models? Refer to the [SQLAlchemy documentation](https://docs.sqlalchemy.org/en/14/orm/basic_relationships.html#one-to-many), try it out, then check out our solution below. There are several ways we _could_ implement a one-to-many relationship in our models. Follow your curiosity if you are interested in some of the other possible approaches. (For example, what is the difference between `back_populates` and `backref`? Is one more efficient?)
+### !callout-info
 
-<br />
+## Defining Relationships as Optional or Required
+
+Just like with regular model properties, when we create relationships between models, we need to carefully consider what makes sense to be optional vs required data. For this project, we're going to make the `Author` relationship optional on the `Book` model, primarily to simplify managing our existing `Book` data that has no author. We are choosing to conceptualize `NULL` author data as representing a book with an unknown author. 
+
```
anselrognlie, line 204:
```suggestion

<br>

```

api-8-relationships-in-sqlalchemy/one-to-many.md
```diff
@@ -55,123 +58,183 @@ An `author` should have the following attributes with the specified types:
...
+If we wanted an `Author` relationship to be required for the `Book` model, we could take a different tactic, such as representing an unknown author by using an "Unknown" author record.
```
anselrognlie, line 205: I think I'd feel more comfortable with an explicit connection made to the `Optional[]` chunk of the syntax. Something along these lines.
```suggestion
If we wanted an `Author` relationship to be required for the `Book` model, we could take a different tactic, such as representing an unknown author by using an "Unknown" author record. In that case, the `Optional[]` portion would be omitted from the mapping declaration. For example, while `Mapped[Optional[int]]` declares a nullable integer column, `Mapped[int]` declares an integer column that disallows `NULL` values, just as Mapped[str] declares a required string column.
```
But I know this would feel weird here, before seein the code. So what about moving the contents of this callout to just after the Book code in the details section?

anselrognlie (changes requested):
Looks good. I just had a follow up thought around the `Optional[]` model declaration.

api-8-relationships-in-sqlalchemy/one-to-many.md
kelsey-steven-ada: Agreed, with the syntax it'll make more sense after they've seen the updated code 👍

api-8-relationships-in-sqlalchemy/one-to-many.md
```diff
@@ -230,24 +323,47 @@ If all else fails, it is okay to delete the migrations directory and the databas
 
 <!-- prettier-ignore-start -->
```
kelsey-steven-ada, line 324:
```suggestion

## Check for Understanding

<!-- prettier-ignore-start -->
```

api-8-relationships-in-sqlalchemy/one-to-many.md
```diff
@@ -198,17 +255,50 @@ class Book(db.Model):
...
+
         return new_book
 ```
+
+### Defining Relationships as Optional or Required
+
+Just like with regular model properties, when we create relationships between models, we need to carefully consider what makes sense to be optional vs required data. For this project, we're going to make the `Author` relationship optional on the `Book` model, primarily to simplify managing our existing `Book` data that has no author. We are choosing to conceptualize `NULL` author data as representing a book with an unknown author. 
+
+<br>
+
```
anselrognlie, line 272: no longer necessary after moving inside details
```suggestion
```

anselrognlie (approved):
Just the last br, then LGTM!

PR #174, opened 2024-09-13
Flask Updates: Building an API pt. 9, lesson 2: Deploying to Render

kelsey-steven-ada:
[Pt. 9 Asana ticket](https://app.asana.com/0/1205655776549324/1206934546187432/f)
[Pt. 9 Lesson 2 ticket](https://app.asana.com/0/1205655776549324/1207017851991917/f)

There is no lesson 1 ticket since that only covers general information about what deployment is; the video there looks at the lesson, but does not show outdated code.

### DIFF
--- a/api-9-deployment/deploying-to-render.md
+++ b/api-9-deployment/deploying-to-render.md
@@ -1,6 +1,7 @@
 # Deploying to Render
 
-<iframe src="https://adaacademy.hosted.panopto.com/Panopto/Pages/Embed.aspx?pid=8285322f-7fe1-45a7-8206-aff701709f35&autoplay=false&offerviewer=true&showtitle=true&showbrand=true&captions=true&interactivity=all" height="405" width="720" style="border: 1px solid #464646;" allowfullscreen allow="autoplay"></iframe>
+<!-- FLASK UPDATE -->
+<!-- <iframe src="https://adaacademy.hosted.panopto.com/Panopto/Pages/Embed.aspx?pid=8285322f-7fe1-45a7-8206-aff701709f35&autoplay=false&offerviewer=true&showtitle=true&showbrand=true&captions=true&interactivity=all" height="405" width="720" style="border: 1px solid #464646;" allowfullscreen allow="autoplay"></iframe> -->
 
 ## Goal
 
@@ -27,7 +28,7 @@ Then, we will cover these topics on continuous deployment to Render:
 
 | Starting Branch | Ending Branch|
 |--|--|
-|`08b-nested-routes` <br> Any branch after connecting the database will work |`09a-deploying-to-render`|
+|`08c-route-utilities-refactors` <br> Any branch after connecting the database will work |`09a-deploying-to-render`|
 
 ### Intro to Render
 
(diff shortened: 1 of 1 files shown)


api-9-deployment/deploying-to-render.md
```diff
@@ -148,6 +149,19 @@ Update the "Start Command" field to `$ gunicorn "app:create_app()"`.
 
 ![Screenshot of changing app start command to gunicorn "app:create_app()"](../assets/deployment/deployment_change-start-command-render.png)
 
+### !callout-info
+
+## Running `gunicorn` locally
+
+If you would like to run your app with `gunicorn` locally from Terminal, this is possible! To do so, we need to provide the required variables (like the  
+`SQLALCHEMY_DATABASE_URI`) as command line arguments, for example :
```
anselrognlie, line 157: I'm worried that calling these command line arguments could be confusing. Could we expand this a bit to explain how environment vairables can be set for the duration of a single command by listing them as name value pairs before the actual command to run?

Would it be worth reminding students that any of the values we've been setting in the `.env` are really being loaded by the `dotenv` library to look as though they are environment variables, but that this doesn't happen when `gunicorn` is hosting our flask app, hence the need to set them ourselves.

There's material along these lines later, but here, it feels like the `SQLALCHEMY_DATABASE_URI` stuff is a little out of the blue. Since it _is_ covered later, maybe than going into all the gory details here, we could remind the reader that this is setting an environment variable and for more details to refer to the "Setup and Initialize the Render Database" topic later on? Or we could move this callout to closer to the configuration section. Or even introduce a new section about trying this out locally as a way to help troubleshoot whether the commands and values we've supplied in the Render configuration are working by running our app under `gunicorn` locally?

api-9-deployment/deploying-to-render.md
```diff
@@ -264,25 +278,32 @@ RENDER_DATABASE_URI = postgresql+psycopg2://YOUR_DATABASE_USERNAME:CONNECTION-ST
...
-db = SQLAlchemy()
-migrate = Migrate()
-
-def create_app(test_config=None):
+def create_app(config=None):
     app = Flask(__name__)
 
     app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
     #app.config['SQLALCHEMY_DATABASE_URI'] = os.environ.get("SQLALCHEMY_DATABASE_URI")
     app.config['SQLALCHEMY_DATABASE_URI'] = os.environ.get("RENDER_DATABASE_URI")
```
anselrognlie, line 295: Totally beyond the scope of this change, but I find it a bit cumbersome that once we introduce the render connection string (and which we need in order to setup up our db tables in the render db, we need to modify our code to go back and forth between the deployed database and the dev database.

What would think about opening a ticket to revisit this approach to make it so that it becomes only a configuration (.env) change? One of the things we stress with deployment in other contexts is that application code should not need to change when moving between development and deployment, but our own example violates this.

api-9-deployment/deploying-to-render.md
```diff
@@ -307,7 +328,7 @@ This will migrate the empty database in our remote Postgres connection to the la
 
 ## Updating the SQLALCHEMY_DATABASE_URI Configuration Key
 
-`SQLALCHEMY_DATABASE_URI` is a [SQLAlchemy Configuration Key](https://flask-sqlalchemy.palletsprojects.com/en/2.x/config/) which allows us to customize certain settings in our app.
+`SQLALCHEMY_DATABASE_URI` is a [SQLAlchemy Configuration Key](https://flask-sqlalchemy.palletsprojects.com/en/3.1.x/config/) which allows us to customize certain settings in our app.
```
anselrognlie, line 331: I was going to suggest using the "latest" version of the doc link (https://flask-sqlalchemy.readthedocs.io/en/latest/config/), but I guess since the version of sqlalchemy we use is fixed to a particular point in time, this could be confusing if folks go read the docs, and try to use features that exist in newer versions, but aren't in the curriculum version. So I guess the specific version of the docs is better?

api-9-deployment/deploying-to-render.md
anselrognlie, line 1: lines 352, 354 give a command to display the book and author tables, but give the command as `d book` and `d author` rather than `\d book` and `\d author`.

anselrognlie (changes requested):
A couple formatting things, a few content thoughts, and some potential future work. Let me know what you think!

api-9-deployment/deploying-to-render.md
kelsey-steven-ada: Since we already have a troubleshooting section, I moved the expanded content down to a subheading under "General Render Debugging Strategies", let me know what you think of the changes/if it fits well there.

api-9-deployment/deploying-to-render.md
kelsey-steven-ada: I prefer that over how we have it now where we make changes in both the `.env` & `__init__.py` files, I'm good with that being added to the backlog.

api-9-deployment/deploying-to-render.md
kelsey-steven-ada: I had a similar thought, it would be nice to have less places we need to make sure we update, but I prefer to err on the side of less potential confusion and frustration for students.

api-9-deployment/deploying-to-render.md
anselrognlie: Yes, I really like the treatment you gave that down in the troubleshooting and love the detailed explanation of what the command actually does. ❤️

anselrognlie (approved):
LGTM!

api-9-deployment/deploying-to-render.md
kelsey-steven-ada: Ticket for that update: https://app.asana.com/0/1201700438643002/1208335052936079

PR #180, opened 2024-10-07
Swaps the order that we refactor to_dict and from_dict to from_dict first

anselrognlie:
On the surface, this just changes the order of the to_dict and from_dict lessons, but in practice, most of both lessons needed to be re-written to adjust what each expected the reader to have already done/encountered. I tried to preserve as much of the original text as I could. The biggest chunk of edits are in `extracting-helper-functions-from-dict.md` (which is now the first of the two) since the justification for starting with that one is much slimmer than starting with to_dict, requiring a lot more build up.

While swapping these, I also noticed that a number of changes from hello-books branch 07b-to-dict-refactor had already been present in 07a-from-dict-refactor, and should be removed. These are detailed in this ticket: https://app.asana.com/0/1205655776549324/1208496538310314/f
But I didn't want to make the changes myself in case there were any considerations to keep in mind due to the interconnectedness of the in-progress branches in that repo.

The Asana tickets for these two lessons are:
- [Building an API pt. 7 - lesson 2: "Extracting Helper Functions Pt. 1" updates](https://app.asana.com/0/1205655776549324/1207017851991907/f)
- [Building an API pt. 7 - lesson 3: "Extracting Helper Functions Pt. 2" updates](https://app.asana.com/0/1205655776549324/1207017851991908/f)

The material added in Pt.1 does have some overlap with the "What is Refactoring" lesson, so there may be some follow on work to see if it makes sense to shift any of this material around, but I would recommend considering that to be external to this review.


api-7-refactoring/extracting-helper-functions-from-dict.md
```diff
@@ -176,35 +332,129 @@ def test_create_one_book_with_extra_keys(client, two_saved_books):
...
+    }
+    return response, 201
+```
+</details>
+
+Now our code handles the `KeyError` when reading from `request_body` by generating a response that tells the client what went wrong, and sets the response status code to 400 Bad Request.
+
+Running our tests, we'll see the two tests that were expecting a `KeyError` to fail, since we're now returning a 400 Bad Request response. The other tests should still pass.
+
+Let's fix the failing tests by updating the expected response status codes and messages. Try updating the tests yourself, then take a look at our updated tests below.
```
yangashley, line 395: Incoming PITA: Is it possible to switch the order of the refactor so that after we show the 500 error in Postman, instead of first changing the route, we write tests and let them fail. Then we show refactoring the route logic (what you have written 357-391)?

Maybe students won't notice but the order as it, to me, feels like the inverse of TDD which I know is regularly done, but want to further reinforce the TDD cycle if possible.

api-7-refactoring/extracting-helper-functions-from-dict.md
```diff
@@ -258,81 +508,164 @@ def test_from_dict_with_extra_keys():
...
 </details>
-</br>
 
-If we run our test suite and see only these new tests are failing, we can start implementing!
+Notice some of the decisions we made in our tests:
+- We're testing that `from_dict` returns a `Book` instance with the correct attributes when given a dictionary with the correct keys.
+- We're testing that `from_dict` raises a `KeyError` when given a dictionary with missing keys.
+- We're testing that `from_dict` ignores extra keys in the dictionary.
+
+These aren't the only behaviors we could have decided on, but they'll meet our initial needs of making sure we have the data we need to create a `Book`. How we communicated that keys are missing was up to us, but since Python raises a `KeyError` when we try to access a key that doesn't exist in a dictionary, we decided to raise a `KeyError` in our tests as well. This will also work well with the existing error logic in the `create_book` route. It would also have been a valid to make a different decision about how to handle extra keys. Ignoring them is a common choice, but we could also have restricted the endpoint to only allow valid keys
```
yangashley, line 517:
```suggestion
These aren't the only behaviors we could have decided on, but they'll meet our initial needs of making sure we have the data we need to create a `Book`. How we communicated that keys are missing was up to us, but since Python raises a `KeyError` when we try to access a key that doesn't exist in a dictionary, we decided to raise a `KeyError` in our tests as well. This will also work well with the existing error logic in the `create_book` route. It would also have been valid to make a different decision about how to handle extra keys. Ignoring them is a common choice, but we could also have restricted the endpoint to only allow valid keys
```

api-7-refactoring/extracting-helper-functions-to-dict.md
```diff
@@ -173,50 +144,49 @@ def test_get_all_books_with_title_query_matching_one(client, two_saved_books):
...
+We know that we want to move our code that creates a dictionary from a `Book` model from the `get_all_books` and `get_one_book` functions, but where does that code belong? Since creating a dictionary from a `Book` requires specific knowledge about the `Book` class, it makes a lot of sense to add the function directly to the `Book` class. 
 
-Planning our function further, we know that we need a book as input, and we want our function's output to be a dictionary with a key for each of the input `Book`'s attributes. Because we need an instance of a `Book` to take this action, we probably want this to be an instance function of our `Book` class.
+Planning our function further, we know that we need a book as input, and we want our function's output to be a dictionary with a key for each of the input `Book`'s attributes. Because we need an instance of a `Book` to take this action, we probably want this to be an instance function of our `Book` class. And since the function will be responsible for converting a `Book` instance to a dictionary, we'll name it `to_dict`.
 
-### Write failing tests for `to_dict`
+### Writing failing tests for `to_dict`
```
yangashley, line 180: Nit: Other titles use capital case (?) for the title. Could we update this to match?
```suggestion
### Writing Failing Tests for `to_dict`
```

api-7-refactoring/extracting-helper-functions-to-dict.md
```diff
@@ -173,50 +144,49 @@ def test_get_all_books_with_title_query_matching_one(client, two_saved_books):
...
 <details>
-   <summary>Complete <code>test_models.py</code> example</summary>
+   <summary>New tests for <code>to_dict</code> in <code>test_book_model.py</code></summary>
 
 ```python
-from app.models.book import Book
-
 def test_to_dict_no_missing_data():
     # Arrange
     test_data = Book(id = 1,
```
yangashley, line 192: While you're in here already, can we make the whitespaces around the named args consistent in this file?

Line 192 and 235 shows white spaces around the equal sign for `id` but then but line 221 correctly omits them

api-7-refactoring/extracting-helper-functions-to-dict.md
```diff
@@ -276,119 +246,207 @@ def test_to_dict_missing_description():
 ```
 
 </details>
-</br>
 
-At this point, when we run our test suite we should see all of the tests in `test_routes.py` passing, and all of the tests in `test_models.py` failing. Now we can start our cycle of making small changes and testing until we complete our refactor!
+At this point, when we run our test suite we should see all of the tests in `test_book_routes.py` passing, and the new tests in `test_book_model.py` failing. Now we can start our cycle of making small changes and testing until we complete our refactor!
 
-### Write the `to_dict` function
+### Writing the `to_dict` function
```
yangashley, line 252:
```suggestion
### Writing the `to_dict` Function
```

api-7-refactoring/extracting-helper-functions-from-dict.md
```diff
@@ -47,35 +54,140 @@ Our test folder should have 2 files:
...
+
+At this point, we might not see much to refactor, so let's keep looking.
+
+The next function is `get_all_books`. This function retrieves all the `Book` instances from the database, applying a number of optional filters, then creates a list of dictionaries, each representing a `Book` instance. The function returns the list of dictionaries as the response body.
+
+This code is definitely more complicated than the `create_book` function. Maybe there's some way we can simplify all that filtering code? It would be great if it could be reused for any other model types that we add to the project! For the moment, let's put a pin in that idea, since there's another possible refactor in this function that we can tackle a little more directly.
+
+`get_all_books` needs to return a collection of `Book` models as a list of dictionaries. Each `Book` model is used to build a dictionary in the response. We can see that the code to build a dictionary from a `Book` model is repeated in the `create_book` function. Taking a quick glance at the next function, `get_one_book`, we can see that it also builds a dictionary from a `Book` model.
```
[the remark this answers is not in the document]
anselrognlie: I was hoping to avoid duplicating the code yet again, but yeah, I can do that. I agree, it's a long chunk of text.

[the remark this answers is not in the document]
yangashley: Ah ok, well maybe in that case it could be in a detail drop down or we can omit them completely.

api-7-refactoring/extracting-helper-functions-to-dict.md
anselrognlie: Our titles are a little all over the place, but I also do tend to prefer having them capitalized.

api-7-refactoring/extracting-helper-functions-to-dict.md
anselrognlie: I'm super on the fence about this. The code here isn't just in learn, it's also in hello-books (this is a direct copy paste). So that spacing issue is also in hello-books. How much do we want to allow the two to diverge (and in every branch)? Should we fix things just in Learn?

api-7-refactoring/extracting-helper-functions-from-dict.md
anselrognlie: Had a discussion on [Slack](https://adadevelopersacademy.slack.com/archives/C01C3J66FM1/p1728426022393249?thread_ts=1728342484.735719&cid=C01C3J66FM1) where it seems that digging into the implications of this change really points more at getting the create error handling code in there earlier so that we don't need to do this mini-refactor at all at this point (it would have been caught with manual postman testing in part 3 when we make the create route to begin with).

So to save effort here (as that feels like a better long term solution), we'll won't fix this and just leave it as is (even if it goes out like this for the current cohort).

api-7-refactoring/extracting-helper-functions-to-dict.md
anselrognlie: Won't fix for now, but opened a ticket under the Update Flask Content for us to discuss as a team: https://app.asana.com/0/1201700438643002/1208496538310317/f

api-7-refactoring/extracting-helper-functions-from-dict.md
anselrognlie: Asana ticket tracking that we should have a discussion about whether to move the inconvenient-to-refactor error handling to an earlier lesson: https://app.asana.com/0/1205655776549324/1208496538310318/f

yangashley (approved):

PR #182, opened 2025-02-05
Update Designing Responses to show returning 204 for successful PUT and DELETE requests

yangashley:
[Include 204 No Content as possible (preferred?) response codes for PUT/DELETE](https://app.asana.com/0/1205655776549324/1208546343899252)

### DIFF
--- a/intro-to-api-design/designing-responses.md
+++ b/intro-to-api-design/designing-responses.md
@@ -206,7 +206,7 @@ Or, of course, Eryn could choose to not send back any data besides a status code
 
 ## Response Bodies Are Optional
 
-RESTful APIs send back self-descriptive responses. Sometimes, an HTTP status code and message is all you need! For example, we can expect to commonly see web APIs send back responses without response bodies for successful `PUT` and `DELETE` requests.
+RESTful APIs send back self-descriptive responses. And sometimes, an HTTP status code is all the description you need! For example, we can expect to commonly see web APIs send back responses without response bodies for successful `PUT` and `DELETE` requests.
 
 ### !end-callout
 
@@ -216,7 +216,7 @@ What should the web API return when a client sends a request to delete a student
 
 - `DELETE` `/students/67`
 
-Eryn could send back a response body with a message that confirms the resource was deleted in conjunction with the [status code `204`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) that indicates the action was successful. However, typically, we do not need to send a response body because the status code `204` is sufficient for conveying that the server successfully fulfilled the request.
+Eryn _could_ send back a response body with a message that confirms the resource was deleted in conjunction with a [status code `200`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200), which indicates the action was successful. However, since a status message such as `"Successfully deleted student 67"` doesn't tell us any more than the `200` (indicating success) itself, we may choose to not include the message. And if we have no message, leaving our response body empty, we can use the [status code `204`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204), which explicitly communicates that the lack of response body was intentional. Status code `204` alone is sufficient for conveying that the server successfully fulfilled the request.
 
 ## Example Outcomes
 
@@ -230,13 +230,26 @@ We can combine meaningful HTTP statuses and intentional response bodies to creat
 | What is the response for _unsuccessfully_ creating a new resource?             | _Invalid_ resource data, such as missing a required attribute     | `400 Bad Request`          | JSON with details about the missing attribute |
 | What is the response for reading the details of a resource?                    | -                                                                 | `200 OK`          | JSON with the resource                        |
 | What is the response for reading the details of a resource that doesn't exist? | -                                                                 | `404 Not Found`   | -                                             |
-| What is the response for updating a resource successfully?                     | Valid resource data                                               | `204 No Content`          | -                                             |
-| What is the response for updating a resource _unsuccessfully_?                 | _Invalid_ resource data, such as violating a data type constraint | `400 Bad Request` | JSON with details about the error             |
+| What is the response for replacing an entire resource successfully?                     | Valid resource data                                               | `204 No Content`          | -                                             |
+| What is the response for replacing an entire resource _unsuccessfully_?                 | _Invalid_ resource data, such as violating a data type constraint | `400 Bad Request` | JSON with details about the error             |
+| What is the response for modifying _some part_ of a resource successfully?                     | Valid resource data                                               | `200 OK`          | JSON with the updated resource                                             |
+| What is the response for modifying _some part_ of a resource _unsuccessfully_?                 | _Invalid_ resource data, such as violating a data type constraint | `400 Bad Request` | JSON with details about the error             |
 | What is the response for updating a resource that doesn't exist?               | Valid resource data                                               | `404 Not Found`   | -                                             |
 | What is the response for deleting a resource successfully?                     | -                                                                 | `204 No Content`          | -                                             |
 | What is the response for deleting a resource that doesn't exist?               | Valid resource data                                               | `404 Not Found`   | -                                             |
 
-## Examples: Slack Web API
+### !callout-info
+
+## Not All APIs Follow RESTful Conventions!
+
+Recall from the previous lesson, "Designing Endpoints", that a central feature of RESTful APIs is that the technical way that a client interacts with the API should be predictable. We talked through many examples of how different requests and responses should look which conform to REST conventions. However, keep in mind that there are API designers that choose to create APIs that do not follow these conventions.
+
+### !end-callout
+
+<details>
+   <summary>Expand to see an example of an API that does not follow the RESTful conventions we outlined in this lesson</summary>
+
+## Slack Web API
 
 Let's consider what the [Slack Web API](https://api.slack.com/web) responds with. Slack describes their web API like so:
 
@@ -318,6 +331,7 @@ An example situation for an unsuccessful response is when the message isn't foun
   "ok": false
 }
 ```
+</details>
 
 ## Check for Understanding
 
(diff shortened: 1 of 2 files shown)


intro-to-api-design/designing-responses.md
```diff
@@ -205,10 +206,18 @@ Or, of course, Eryn could choose to not send back any data besides a status code
...
 
 ### !end-callout
 
+### Example: `DELETE` `/students/67` 
+
+What should the web API return when a client sends a request to delete a student resource that exists? Imagine a client sending this request: 
+
+- `DELETE` `/students/67`
+
+Eryn could send back a response body with a message that confirms the resource was deleted in conjunction with the [status code `204`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) that indicates the action was successful. However, typically, we do not need to send a response body because the status code `204` is sufficient for conveying that the server successfully fulfilled the request.
```
yangashley, line 219: The [task](https://app.asana.com/0/1205655776549324/1208546343899252) says "Marked medium scale in case we feel the need to expand on why. If we just add it as a possible response in the table of responses, this should be small."

I did introduce 204 in the tables, but also tried to explain why. How do you feel about the addition to the callout with this example?

anselrognlie: Here are some of the things that come to mind for me when discussing this topic.

- 200 + dictionary of deleted record: If we're deleting it, why do we need to send the data back? Most probably, a user was in a situation where they were already looking at the record (so the app already had the data for the record) and they took an action (like clicking a button) to remove the item. Even if we _didn't_ return the data for the deleted item, if the app really needed to fetch it, it could do an initial GET before the DELETE.
- 200 + successfully deleted message: A message like "successfully deleted record 1" doesn't provide much benefit. We already know the deletion was successful from the status code, and the message itself isn't really structured for computer use. We should remember that end users don't typically interact with API calls, so they won't see the response. Instead, users interact with an app (code) that makes the API call and can display its own success or failure messages.
- 204 with no response body: The 204 indicates the request was successful, and that there should be no attempt to read the response body (because there isn't one). From the other two cases, what would make sense to be returned. Neither the deleted record data, nor a status message is immediately useful, so leaving it off entirely works well.

The same scenarios apply to PUT. With PUT, we _might_ still choose to return the updated record, but since we must send all of the data for the updated record, then in theory we have all the data we need to update any local data ourselves if necessary (though there's a separation of concerns argument that could be made to tilt things in favor of returning the updated record).

PATCHes are more likely to return the updated record, since we may have only sent a small piece of the entire record, or even made use of a custom route that performs additional work on the record in ways the app wouldn't necessarily know about.

So we could make a recommendation that 200 + the record in the response could be used for all cases (even DELETE), 200 + some status message is pretty useless (though error messages can be helpful on failures), and if the record dictionary really isn't needed, then 204 with empty response can be suitable.

anselrognlie, line 219: It's probably sufficient to keep things more focused as you have, but in that case, it reads a bit confusing to me that we talk about combining 204 with a message. Can we start from 200 with a message, then drop the message, and then switch the 200 to 204? Something like this?
```suggestion

Eryn _could_ send back a response body with a message that confirms the resource was deleted in conjunction with a [status code `200`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200), which indicates the action was successful. However, since a status message such as `"Successfully deleted student 67"` doesn't tell us any more than the `200` (indicating success) itself, we may choose to not include the message. And if we have no message, leaving our response body empty, we can use the [status code `204`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204), which explicitly communicates that the lack of response body was intentional. Status code `204` alone is sufficient for conveying that the server successfully fulfilled the request.
```

intro-to-api-design/designing-responses.md
```diff
@@ -221,10 +230,10 @@ We can combine meaningful HTTP statuses and intentional response bodies to creat
 | What is the response for _unsuccessfully_ creating a new resource?             | _Invalid_ resource data, such as missing a required attribute     | `400 Bad Request`          | JSON with details about the missing attribute |
 | What is the response for reading the details of a resource?                    | -                                                                 | `200 OK`          | JSON with the resource                        |
 | What is the response for reading the details of a resource that doesn't exist? | -                                                                 | `404 Not Found`   | -                                             |
-| What is the response for updating a resource successfully?                     | Valid resource data                                               | `200 OK`          | -                                             |
+| What is the response for updating a resource successfully?                     | Valid resource data                                               | `204 No Content`          | -                                             |
```
anselrognlie, line 233: No response body for 204. Or I'd be open to have two options for this. 200 with the resource data, or 204 with no response.
```suggestion
| What is the response for updating a resource successfully?                     | -                                               | `204 No Content`          | -                                             |
```

anselrognlie: Should we make any kind of distinction between a PUT update (replace) and a PATCH update (modify)? From the thoughts I left above, it does feel resonably common to return the record on a PATCH (and we do it with custom routes for like marking tasks complete or the pet counts in Flasky).

intro-to-api-design/designing-responses.md
```diff
@@ -221,10 +230,10 @@ We can combine meaningful HTTP statuses and intentional response bodies to creat
...
 | What is the response for updating a resource _unsuccessfully_?                 | _Invalid_ resource data, such as violating a data type constraint | `400 Bad Request` | JSON with details about the error             |
 | What is the response for updating a resource that doesn't exist?               | Valid resource data                                               | `404 Not Found`   | -                                             |
-| What is the response for deleting a resource successfully?                     | -                                                                 | `200 OK`          | -                                             |
+| What is the response for deleting a resource successfully?                     | -                                                                 | `204 No Content`          | -                                             |
 | What is the response for deleting a resource that doesn't exist?               | Valid resource data                                               | `404 Not Found`   | -                                             |
 
 ## Examples: Slack Web API
```
anselrognlie, line 239: Beyond this, we show examples from the slack api for posting a message and deleting a message. The delete contradicts our own example, since Slack is an HTTP-RPC API rather than a RESTful API.

Should we either:
1) (drastic) cut the Slack API examples, since they undercut our material
2) Add some very loud reminders that the Slack API is not RESTful (but then, why are we showing it in this context?)

anselrognlie (changes requested):
The main parts of the change looks good (a few comments here and there), but I think there's a few other things we might want to consider looking at while we're here.

intro-to-api-design/designing-responses.md
yangashley: I'm trying to decide if it's helpful for students to see an example of how APIs can differ even though there are conventions or if it's a lot of information all at once and instead of being helpful might be confusing. 

What if we add a callout that says something like "not all APIs are RESTful" and then put the Slack example walk though on lines 239-320 in a detail "Expand to see an example of an API that doesn't follow the conventions we've outlined" ?

intro-to-api-design/designing-responses.md
yangashley: Do you mean like adding additional entries to the table?

What about something like this (what I added is highlighted in red). I tried to make the distinction between PUT and PATCH, but if this doesn't feel clear enough I can explicitly add the verbs to the examples.

image

intro-to-api-design/designing-responses.md
yangashley: Thanks for this, I'm ruminating over how I think this should be explained and how to format it.

anselrognlie: I don't know that this all needs to be covered now (or at all). Maybe we could just capture this in an instructor doc as some of the concepts behind why we made certain recommendations? The DELETE example does a good job of introducing some of this nuance as is.

intro-to-api-design/designing-responses.md
anselrognlie: That sounds good to me as is, though I agree, it might not be clear how those differ in terms of verb. But I bet if a student asked about it in Slack, another student would be able to answer them. I'm mostly thinking that it would be nice to have justification in the text to point back at when we do the PATCH custom routes later on, without bringing up custom routes now (since they also muddy the RESTful route model). This feels like it fits the bill.

We could try using the same phrasing the Designing endpoints lesson uses. It terms PUT as "replacing a resource" and PATCH as "updating a resource", and also talks about "part of an updated resource".

So maybe instead of "updating an entire resource" we could use "replacing an entire resource"? The PATCH one seems clear and similar enough to me as is.

intro-to-api-design/designing-responses.md
anselrognlie: Oooh. I love that!

intro-to-api-design/intro-to-api-design.instructor.md
```diff
@@ -13,6 +13,22 @@ Notes for teaching:
...
+
+The same scenarios apply for PUT responses. With PUT, we might still choose to return the updated record, but since we must send all of the data for the updated record, then in theory we have all the data we need to update any local data ourselves if necessary (though there's a separation of concerns argument that could be made to tilt things in favor of returning the updated record).
+
+PATCH responses are more likely to return the updated record, since we may have only sent a small piece of the entire record, or even made use of a custom route that performs additional work on the record in ways the app wouldn't necessarily know about.
+
+We can explain that:
+- Returning status code 200 and the record in the response _could_ be used for all cases (even DELETE)
+- Returning status code 200 with some status message isn't useful
+- Returning status code 204 with an empty response is suitable if the record dictionary isn't needed
+
```
yangashley, line 31: I took what you wrote in your comment and re-formatted it a little bit and added it here to the instructor doc.

anselrognlie: Ha ha. I didn't realize there was already an instructor doc there.

anselrognlie (approved):
LGTM. I think hiding the Slack stuff help focuses this and feels less noisy. It's there if they want to look at ti, but now it's very clear this is not the pattern we want them to follow.

PR #183, opened 2025-02-21
Adds missing semicolons throughout Intro to SQL

anselrognlie:
Work started in response to student filing about missing semicolons in the deleting lesson.

I went through the whole Intro to SQL to check for any other missing semicolons, and reorganized the syntax table explanations to highlight the inclusion of the semicolon. Previously the semicolon was just stuck to the end of the last chunk of syntax, which was sometimes an optional chunk. Every example now consistently calls out that the semicolon terminates the SQL query.

Asana: https://app.asana.com/0/1205655776549324/1204482550755381


kelsey-steven-ada (approved):
Changes look great, thanks for updating the tables in the lessons to call out directly the semicolons terminating statements!

PR #185, opened 2025-08-06
Update instructions for installing the requests package

yangashley:
[Task](https://app.asana.com/1/181459410160484/project/1205655776549324/task/1210193230464615?focus=true): Refresh instructions for installing requests

### DIFF
--- a/requests-and-responses/sending-requests-with-python.md
+++ b/requests-and-responses/sending-requests-with-python.md
@@ -13,9 +13,9 @@ Our goal for this lesson is to:
 
 Ebony is extremely interested in history. She takes notes on all of the landmarks in the world.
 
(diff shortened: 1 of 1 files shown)


requests-and-responses/sending-requests-with-python.md
```diff
@@ -52,10 +52,12 @@ When we take a minute to look through the [`requests` package documentation](htt
 
 ... and so many more explanations and examples about how to use the package!
 
-We can install `requests` with this command:
+Recall that when we work on Python projects, even small ones like Ebony's New 7 Wonders, we should first create a virtual environment and then activate it before installing packages.
```
yangashley, line 55: The task says "It would also be helpful to remind the reader (perhaps a reflection by Ebony, the avatar in the lesson) that this needs to be done in a venv."

How does this sound to you?

kelsey-steven-ada: This sounds good to me, we're in unit 2 so I think the reminder is good without explicitly writing out the `venv` commands too.

kelsey-steven-ada (approved):
LGTM!

PR #187, opened 2025-08-25
Building an API pt. 7 - Replace tests for update & delete refactors

kelsey-steven-ada:
Asana Task

Changes
- Update the tests for update and delete routes in "BaA Pt. 7" > "Refactoring for Future Features" lesson to use current api pattern where update and delete do not return a response body.


mikellewade (approved):
LGTM ✅

PR #189, opened 2025-08-27
BaA Pt. 1 > Hello World Routes - Include setup instructions for debugging Flask through VS Code

kelsey-steven-ada:
[Asana Task](https://app.asana.com/1/181459410160484/project/1205655776549324/task/1202242375423578?focus=true)

Changes
- Reorganize "Troubleshooting" information in "Building an API Pt. 1" > "Hello World Routes" and add link to new resource on debugging a Flask application
- Create resource doc for debugging a Flask app that covers creating a launch.json file and using it to run the debugger
- Update config.yaml to include new doc as a resource file so formatting renders


api-1-setup-read/hello-world-routes.md
```diff
@@ -152,24 +152,28 @@ Instead of using Postman, we can alternatively use our browser to make a `GET` r
...
 
-Here are a couple quick troubleshooting techniques to try:
+<!-- available callout types: info, success, warning, danger, secondary, star  -->
+### !callout-info
 
-</br>
+## Debugging a Flask Application
 
-If you are unable to access the server at `localhost:5000/`, replace `localhost:5000` with `127.0.0.1` (i.e. `127.0.0.1:5000/`)
+We recommend debugging Flask from tests once we get to test set up materials, since there is a little extra set up involved to debug a Flask application on its own. If you get stuck in the meantime, we can debug a Flask application using [this doc for reference](debug-flask-app-supplement.md).
```
anselrognlie: Totally agree with the sentiments here, but it felt like this could also provide an opportunity to more explicitly reinforce debugging through tests as our preferred approach generally. How about something like this?
```suggestion
As with other code we have written, we can use the VS Code debugger to help us troubleshoot code in our Flask applications. Once we reach the material on writing tests for Flask, we'll see that the process of debugging Flask through tests is identical to how we've used tests for debugging in our previous work, requiring no additional setup.

In the meantime, if we encounter problems that we're having trouble resolving using alternative debugging techniques or the strategies below, we can use the following guide for a brief introduction on using the debugger with Flask applications directly.

- [Debugging a Flask Application](debug-flask-app-supplement.md)
```

api-1-setup-read/hello-world-routes.md
```diff
@@ -152,24 +152,28 @@ Instead of using Postman, we can alternatively use our browser to make a `GET` r
...
-</br>
+### !end-callout
 
-If you receive the error “Access to localhost was denied”, go to **System Preferences** -> **Sharing**, and uncheck “AirPlay Receiver”.
+### Troubleshooting Reaching the Local Server
 
-### !end-callout
+Here are a couple quick troubleshooting techniques to try if you are having trouble reaching your local server:
+- If you are unable to access the server at `localhost:5000/`, replace `localhost:5000` with `127.0.0.1` (i.e. `127.0.0.1:5000/`)
+- If you receive the error “Access to localhost was denied”, go to **System Preferences** -> **Sharing**, and uncheck “AirPlay Receiver”.
```
anselrognlie, line 178: Any bets on how many folks still run into this? 😭

api-1-setup-read/debug-flask-app-supplement.md
```diff
@@ -0,0 +1,76 @@
+# Debugging a Flask Application
+
+To set breakpoints and debug a Flask app without tests we need to:
+1. Create a debug launch configuration for a Flask application in VS Code
+2. Run the app using the new launch configuration
+3. Add breakpoints
+4. Use postman to hit the endpoints where we've added our breakpoints
```
anselrognlie: Capitalization, and minor word choice
```suggestion
4. Use Postman to execute the endpoints where we've added our breakpoints
```

api-1-setup-read/debug-flask-app-supplement.md
```diff
@@ -0,0 +1,76 @@
...
+
+This guide will cover the first 2 steps to get us up and running!
+
+## Create a Debug Launch Configuration
+
+We need a way to tell VS Code to launch our Flask app in debug mode rather than trying to run a routes file as a standalone Python file. 
+- To do so, we'll have VS Code help us generate a `launch.json` file that instructs VS Code to start up our app in debug mode.
```
anselrognlie: For some additional justification, this felt like a good opportunity to bring up the parallel reason for why debugging tests themselves is handled a little differently in VS Code.
```suggestion
## Debugging Flask Requires Additional Configuration

Let's think back to when we started using tests to debug our code. A common place we ran into trouble was if we tried to debug our tests directly using the Run and Debug panel. Because of the typical way our tests import our project code, doing so usually resulted in an error—an `ImportError`. In fact, we would have seen the same import error if we had tried to run our test file in the terminal using a command like `python tests/test_file.py`. To properly run our tests, we didn't run the `python` command, we ran `pytest`.

When debugging, this meant that we couldn't use the Play button—even though it appears in several places in VS Code—to start debugging our tests. Instead, we needed VS Code to start the debugger on our tests in a similar way to how `pytest` starts our tests in the terminal. To do this, we debugged our tests through the Test Explorer panel.

We find ourselves in a similar situation if we try to debug a Flask application. When we run our Flask application, we don't use a command like `python app/__init__.py`. Instead, we start our Flask application with the `flask` command (sometimes with additional arguments).

So in a similar fashion to how we needed to figure out the VS Code debugging equivalent of running `pytest`, we now need to figure out the VS Code debugging equivalent of running `flask`.

- To do so, we'll have VS Code help us generate a `launch.json` file that instructs VS Code to start up our Flask application in debug mode.

## Create a Debug Launch Configuration
```

api-1-setup-read/debug-flask-app-supplement.md
```diff
@@ -0,0 +1,76 @@
...
+    ![VS Code UI showing a dropdown from the code window that has the option "More Python Debugger options..." highlighted](assets/debug_flask_without_tests/add_config_2_more_python_options.png)
+    _Fig. VS Code's UI showing a dropdown from the code window that has the option "More Python Debugger options..." highlighted. ([Full size image](assets/debug_flask_without_tests/add_config_2_more_python_options.png))_
+
+3. A new menu will appear, and we want to press the option: "Python Debugger: Flask hello-books-api".
+    ![VS Code UI showing a dropdown from the code window displaying all Python Debugger options with an arrow pointing to "Python Debugger: Flask hello-books-api"](assets/debug_flask_without_tests/add_config_3_python_debugger_flask.png)
+    _Fig. VS Code's UI showing a dropdown from the code window with the option "Python Debugger: Flask hello-books-api" highlighted. ([Full size image](assets/debug_flask_without_tests/add_config_3_python_debugger_flask.png))_
+
+4. After step 3, we will be shown a new file with the launch.json that VS Code generated for us. We can safely close this file once it is created.
```
anselrognlie:
```suggestion
4. After step 3, we will be shown a new file with the `launch.json` that VS Code generated for us. We can safely close this file once it is created.
```

api-1-setup-read/debug-flask-app-supplement.md
```diff
@@ -0,0 +1,76 @@
...
+    ![VS Code UI showing the generated launch.json file](assets/debug_flask_without_tests/add_config_4_generated_flask_debug_configuration.png)
+    _Fig. The generated launch.json file. ([Full size image](assets/debug_flask_without_tests/add_config_4_generated_flask_debug_configuration.png))_
```
anselrognlie:
```suggestion
    _Fig. The generated `launch.json` file. ([Full size image](assets/debug_flask_without_tests/add_config_4_generated_flask_debug_configuration.png))_
```

api-1-setup-read/debug-flask-app-supplement.md
```diff
@@ -0,0 +1,76 @@
...
+
+    The full text of the launch.json file we generated is below:
```
anselrognlie:
```suggestion
    The full text of the `launch.json` file we generated is below:
```

api-1-setup-read/debug-flask-app-supplement.md
```diff
@@ -0,0 +1,76 @@
...
+
+2. In the drop down that appears, click "Python Debugger: Flask".
+    ![VS Code UI showing a drop down menu with the option "Python Debugger: Flask" highlighted](assets/debug_flask_without_tests/run_app_2_choose_python_debugger_flask.png)
+    _Fig. VS Code UI showing a drop down menu with the option "Python Debugger: Flask" highlighted. ([Full size image](assets/debug_flask_without_tests/run_app_2_choose_python_debugger_flask.png))_
+
+At this point VS Code should start running the application in debug mode in a terminal. 
+![VS Code with the debug UI at the top of the window and the Flask application running in a terminal in a panel on the bottom of the window.](assets/debug_flask_without_tests/run_app_3_flask_app_running_in_debug_with_breakpoint.png)
+_Fig. VS Code with the debug UI at the top of the window and the Flask application running in a terminal. ([Full size image](assets/debug_flask_without_tests/run_app_3_flask_app_running_in_debug_with_breakpoint.png))_
+
+From here we can set breakpoints in our route functions then use Postman to send requests to trigger the endpoints we want to debug!
```
anselrognlie: While I think we don't need to go into a specific example, a little more context, and then a summary to reiterate what we did feels helpful to me.
```suggestion
From here we can set breakpoints in our route functions, then use Postman to send requests to trigger the endpoints we want to debug!

A great way to think about where to place a breakpoint is to consider where we might like to place a `print()` statement, then place a breakpoint there instead.

We must also remember that it's not enough to just start the debugger, even after setting breakpoints. In order for the debugger to be able to stop at a breakpoint, we need to trigger the code with the breakpoint to actually run. When Flask is started under the debugger, all it does is wait for incoming requests, the same as when we run `flask` from the terminal. This is why we need to use Postman to send a request. The incoming request will cause the related endpoint function to run, allowing any breakpoints we set in that endpoint (or other code called from that endpoint) to be triggered.

## Summary

While we will often prefer to debug our Flask applications by using our tests to exercise our code, there are additional concerns around writing Flask tests that mean we're not yet ready to tackle writing Flask tests. In the meantime, if we run into problems with our Flask application that we're finding challenging to investigate with non-debugger approaches (such as visually inspecting our code, reviewing the logs, or rubber ducking), it's possible to set up VS Code to be able to debug our Flask application.

To debug our Flask application with VS Code, we
1. Create a Launch Configuration to tell VS Code how to start our Flask application.
2. Start the debugger using the Flask Launch Configuration.
3. Place breakpoints in our code at points of interest.
4. Send requests to the Flask application using Postman that will cause the code with the breakpoints to run.
```

api-1-setup-read/debug-flask-app-supplement.md
```diff
@@ -0,0 +1,76 @@
...
+4. Use postman to hit the endpoints where we've added our breakpoints
+
+This guide will cover the first 2 steps to get us up and running!
+
+## Create a Debug Launch Configuration
+
+We need a way to tell VS Code to launch our Flask app in debug mode rather than trying to run a routes file as a standalone Python file. 
+- To do so, we'll have VS Code help us generate a `launch.json` file that instructs VS Code to start up our app in debug mode.
+
+1. To create a new launch configuration in VS Code, press the "Run" menu at the top of the screen and choose the option "Add Configuration".
```
anselrognlie, line 25: I take a slightly different path to setting up the debugger, but it looks like they end up in the same place.

I like that both here, as well as for running the debugger below, your approach avoids the Run and Debug panel. Hopefully this will help the reader keep this setup separate from the general debug flow in their minds.

anselrognlie (changes requested):
Looks good overall, and I like the specific approach you took for the setup and debugging. Minor suggestions here and there, as well as some areas I felt might benefit from providing some additional background.

api-1-setup-read/debug-flask-app-supplement.md
```diff
@@ -0,0 +1,99 @@
...
+_Fig. VS Code with the debug UI at the top of the window and the Flask application running in a terminal. ([Full size image](assets/debug_flask_without_tests/run_app_3_flask_app_running_in_debug_with_breakpoint.png))_
+
+From here we can set breakpoints in our route functions, then use Postman to send requests to trigger the endpoints we want to debug!
+
+A great way to think about where to place a breakpoint is to consider where we might like to place a `print()` statement, then place a breakpoint there instead.
+
+We must also remember that it's not enough to just start the debugger, even after setting breakpoints. In order for the debugger to be able to stop at a breakpoint, we need to trigger the code with the breakpoint to actually run. When Flask is started under the debugger, all it does is wait for incoming requests, the same as when we run `flask` from the terminal. This is why we need to use Postman to send a request. The incoming request will cause the related endpoint function to run, allowing any breakpoints we set in that endpoint (or other code called from that endpoint) to be triggered.
```
kelsey-steven-ada:
```suggestion
From here we can set breakpoints in our route functions, then use Postman to send requests to trigger the endpoints we want to debug!
- A great way to think about where to place a breakpoint is to consider where we might like to place a `print()` statement, then place a breakpoint there instead.

We must also remember that it's not enough to just start the debugger, even after setting breakpoints. In order for the debugger to be able to stop at a breakpoint, we need to trigger the code with the breakpoint to actually run. When Flask is started under the debugger, all it does is wait for incoming requests, the same as when we run `flask` from the terminal. 
- This is why we need to use Postman to send a request. The incoming request will cause the related endpoint function to run, allowing any breakpoints we set in that endpoint (or other code called from that endpoint) to be triggered.
```

api-1-setup-read/debug-flask-app-supplement.md
```diff
@@ -0,0 +1,99 @@
...
+    _Fig. VS Code UI showing a drop down menu with the option "Python Debugger: Flask" highlighted. ([Full size image](assets/debug_flask_without_tests/run_app_2_choose_python_debugger_flask.png))_
+
+At this point VS Code should start running the application in debug mode in a terminal. 
+![VS Code with the debug UI at the top of the window and the Flask application running in a terminal in a panel on the bottom of the window.](assets/debug_flask_without_tests/run_app_3_flask_app_running_in_debug_with_breakpoint.png)
+_Fig. VS Code with the debug UI at the top of the window and the Flask application running in a terminal. ([Full size image](assets/debug_flask_without_tests/run_app_3_flask_app_running_in_debug_with_breakpoint.png))_
+
+From here we can set breakpoints in our route functions, then use Postman to send requests to trigger the endpoints we want to debug!
+- A great way to think about where to place a breakpoint is to consider where we might like to place a `print()` statement, then place a breakpoint there instead.
```
anselrognlie, line 86: I like that you pulled out the more "actionable" stuff with a bullet. Though I really wish Learn markdown had something more like an aside. A callout would be too heavy for these. I think we've used the blockquote `>` for this in some places, but not consistently.

api-1-setup-read/debug-flask-app-supplement.md
```diff
@@ -0,0 +1,99 @@
...
+
+We must also remember that it's not enough to just start the debugger, even after setting breakpoints. In order for the debugger to be able to stop at a breakpoint, we need to trigger the code with the breakpoint to actually run. When Flask is started under the debugger, all it does is wait for incoming requests, the same as when we run `flask` from the terminal. 
+- This is why we need to use Postman to send a request. The incoming request will cause the related endpoint function to run, allowing any breakpoints we set in that endpoint (or other code called from that endpoint) to be triggered.
+
+## Summary
+
+While we will often prefer to debug our Flask applications by using our tests to exercise our code, there are additional concerns around writing Flask tests that mean we're not yet ready to tackle writing Flask tests. In the meantime, if we run into problems with our Flask application that we're finding challenging to investigate with non-debugger approaches (such as visually inspecting our code, reviewing the logs, or rubber ducking), it's possible to set up VS Code to be able to debug our Flask application.
```
anselrognlie: Sorry for this bit of redundancy in my suggestion.
```suggestion
While we will often prefer to debug our Flask applications by using our tests to exercise our code, there are additional concerns around writing Flask tests that mean we're not yet ready to tackle this just yet. In the meantime, if we run into problems with our Flask application that we're finding challenging to investigate with non-debugger approaches (such as visually inspecting our code, reviewing the logs, or rubber ducking), it's possible to set up VS Code to be able to debug our Flask application.
```

anselrognlie (approved):
LGTM. Just noticed a bit of repetitive phrasing in one of my previous suggestions.

PR #190, opened 2025-09-04
Add callout to turn off warnings that Optional[] is deprecated

kelsey-steven-ada:
[Asana Ticket](https://app.asana.com/1/181459410160484/project/1205655776549324/task/1208752650450571?focus=true)

Changes
- Add note about other syntax for optionality in lesson 3 of BaA. 
    - I'm open to removing what I added here, but it felt weird to say we _need_ to use `Optional` to declare nullability when there is other valid syntax.
- Add content to "Fixing type errors" callout to also handle optional warnings in BaA pt. 8


apradoada (approved):
Looks good to me! The image used is super clear!

